PR #13, opened 2015-03-21
Update README.md

yoimox:
add download link


ashimokawa:
I was thinking about something similar but without hotlinking a resource that I have no control over.
I might add an f-droid badge image to the repository and then link that instead.

yoimox:
That's a smart idea.

Issue #15, opened 2015-03-31
Initial MI Band support


danielegobbetti:
Hello there,

I just received my new miband (which I bought only because I've read on fdroid that you were planning to support it :-) ). I just wanted to let you know that I am willing to help implementing its support. BLE is completely new to me, I am reading the sources of https://github.com/paulgavrikov/xiaomi-miband-android to understand the basics.

I guess that the first step to support miband would be to define a specific protocol ( app/src/main/java/nodomain/freeyourgadget/gadgetbridge/protocol ), right?

cpfeiffer:
Hi Daniele,

yes, this is a good start. You would certainly need something like the BluetoothCommunicationService for BLE. There's a few other projects to look for inspiration:
https://github.com/lwis/miband-notifier
https://github.com/UgoRaffaele/miband-notifier
https://bitbucket.org/OscarAcena/mibanda

cpfeiffer:
BTW: looking forward to your miband support -- I actually want(ed) to do that myself, but got sidetracked so far.

ashimokawa:
Hi Daniele,

currently device abstraction is in a bad shape. Since Carsten got sidetracked and I only have a pebble and wanted to have that working fast, some things are a mess.

I suggest you should wait a few day until I cleaned things up a bit. My plan is as follows:

1. Create some abstract base class for the protocols and then have both PebbleProtocol.java and the to-be-created MibandProtocol.java implent that. I will commit a stubbed class for you to work on.
2. Depending on the device type instantiate the corresponding class in BluetoothCommunicationService.java 
3. Create an alternative Miband specific main loop in BluetoothCommunicationService.java.
You would have to create something like BtSocketIoThread for BLE/Miband. Actually BtSocketIoThread is Pebble specific.  I cant do that, since I dont have a Miband and zero knowledge about BLE. It is Carstens or your turn there. All I can do is create a base class and stub the Miband part for you.

To make Gadgetbridges current architecture a bit clearer here is what happens when when an SMS arrives. This is also true for everything else that is send to the Device (emails, music playback information etc)
1. SMSReceiver gets the SMS and calls the BluetoothCommunicationService with the SMS details. This is device independent.
2. The BluetoothCommunicationService gets the SMS and calls PebbleProtocol.encodeSMS(). This will become somthing like deviceProtocol.encodeSMS() which is either an instance of PebbleProtocol or MibandProtocol.
[...]

PR #27, opened 2015-04-24
Implement setting the time on the miband.

danielegobbetti:
It works from the debug activity, since I am not sure about the value of
the ts argument, it always sets the current time.

Created pull request as as asked in https://github.com/Freeyourgadget/Gadgetbridge/issues/15


cpfeiffer:
Cool, thanks :-)

ashimokawa:
Yes, -1 means "current time",
I know it is a bit stupid to replicate code for every protocol implementation to get the current system time, but then again it is the most accurate if we get it as late as possible ;)

PR #58, opened 2015-06-01
Refactoring of the data ingestion method.

danielegobbetti:
* no more ByteBuffer, but a fixed size byte array that gets flushed everytime it's needed
* log of activity data to a separate file (no DB integration yet)
* the size of the buffer must be a multiple of 3 (1 minute of data = 3 bytes) and 20 (the normal size of the chunks we get from the device)
* better logging and more comments in code

### DIFF
--- a/app/src/main/assets/logback.xml
+++ b/app/src/main/assets/logback.xml
@@ -28,6 +28,30 @@
         </encoder>
     </appender>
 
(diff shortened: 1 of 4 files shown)


danielegobbetti:
Please note that the logic of the Logger initialization has changed: activity data are always logged to file at the moment.
Consider this before merging to master.

PR #59, opened 2015-06-02
First WIP data display for the mi band

danielegobbetti:
Added logic to display the data from the miband. It's just an initial stub.

@ashimokawa since I moved the code for the pebble around, please check that I did not broke anything.


ashimokawa:
Judging from a quick look you have that switch in the inner loop although device type does not change.
Could you change that or should I commit something based on your code?

danielegobbetti:
Please do, I prefer if you refactor the class to your likings, I can then personalize the miband data output. Thanks!

ashimokawa:
Thanks, I unified our code.

PR #62, opened 2015-06-04
Get rid of the unused ID in the DB, this could help in:

danielegobbetti:
* reducing the used space
* reducing the time needed for inserts

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/database/ActivityDatabaseHandler.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/database/ActivityDatabaseHandler.java
@@ -12,13 +12,12 @@ import nodomain.freeyourgadget.gadgetbridge.GBActivitySample;
 
 public class ActivityDatabaseHandler extends SQLiteOpenHelper {
 
-    private static final int DATABASE_VERSION = 3;
+    private static final int DATABASE_VERSION = 4;
 
     private static final String DATABASE_NAME = "ActivityDatabase";
 
     private static final String TABLE_GBACTIVITYSAMPLES = "GBActivitySamples";
 
-    private static final String KEY_ID = "id";
     private static final String KEY_TIMESTAMP = "timestamp";
     private static final String KEY_PROVIDER = "provider";
     private static final String KEY_INTENSITY = "intensity";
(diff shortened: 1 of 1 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/database/ActivityDatabaseHandler.java
```diff
@@ -31,22 +30,35 @@ public ActivityDatabaseHandler(Context context) {
...
-        String CREATE_GBACTIVITYSAMPLES_TABLE = "CREATE TABLE " + TABLE_GBACTIVITYSAMPLES + "("
-                + KEY_ID + " INTEGER PRIMARY KEY,"
+        String CREATE_GBACTIVITYSAMPLES_TABLE = "CREATE TABLE " + TABLE_GBACTIVITYSAMPLES + " ("
                 + KEY_TIMESTAMP + " INT,"
                 + KEY_PROVIDER + " TINYINT,"
                 + KEY_INTENSITY + " SMALLINT,"
                 + KEY_STEPS + " TINYINT,"
                 + KEY_TYPE + " TINYINT,"
-                + " UNIQUE (" + KEY_TIMESTAMP + "," + KEY_PROVIDER + ") ON CONFLICT REPLACE)";
+                + " PRIMARY_KEY (" + KEY_TIMESTAMP + "," + KEY_PROVIDER + ") ON CONFLICT REPLACE)";
```
cpfeiffer: A further optimization would be to add WITHOUT_ROWID to the CREATE TABLE statement. That would only work with Lollipop and upwards.

https://sqlite.org/withoutrowid.html
http://stackoverflow.com/questions/2421189/version-of-sqlite-used-in-android

cpfeiffer:
For reference, I just tried it and here's my results:


Version                      | DB file size
-----------------------------|----------------
Original                       | 148679 bytes
Yours                          | 144449 bytes
Yours + without rowid | 78262 bytes

So, at least the file size pretty much halves using WITHOUT ROWID, just as the sqlite docs say.
I didn't test performance with it, though.

cpfeiffer:
I'm wondering if the opening and closing of the sqlite database for every single sample is not a performance problem as well. See ActivityDatabaseHandler.addGBActivitySample().

PR #64, opened 2015-06-07
Add Turkish translation

mtsekm:
Andi, here is my attempt at a Turkish translation. Let me know if it needs further work.


cpfeiffer:
Cool, thanks a lot :-)

ashimokawa:
Thanks :+1: 

You could also use transifex to update the translation. If @cpfeiffer still maintains this.

cpfeiffer:
@ashimokawa yes, I do -- just updated the translations from transifex

PR #66, opened 2015-06-09
Pebble settings should not be in the MiBand specific settings?

chrisvella:
I assume pebble message vibration settings should not be available in the miband specific settings?


ashimokawa:
Actually this is correct.
This is the vibration count for "Pebble Messages".
The term "Pebble Messages" is misleading but I do not have a better idea.
"Pebble Messages" are messages that are sent from external android application explicitly to the Pebble via Intent. We use this for Conversations.
Technically this is not Pebble specific at all - so it is possible to have the Mi Band vibrate or display the message on another Gadgetbridge supported smartwatch in the future.

Any idea how to stop the confusion?

chrisvella:
Fair enough. My mistake.

"3rd party pebble messages", "pebble specific messages" and "device specific messages" are the only replacement names I can think of. I know they are pretty wordy...

cpfeiffer:
I for one would actually use the same configuration as generic notifications, because a user cannot distinguish them anyway. And the configuration option really is misleading.

PR #115, opened 2015-09-12
Improve Pebble Time notifications icons

roidelapluie:
* Add a list of icons in the source code as I discovered them (invisible
  for the user)
* Generic notification now use the correct icon instead of the SMS icon

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java
@@ -70,10 +70,104 @@ public class PebbleProtocol extends GBDeviceProtocol {
     static final byte BLOBDB_REMINDER = 3;
     static final byte BLOBDB_NOTIFICATION = 4;
 
+    // This is not in the Pebble protocol
+    static final byte NOTIFICATION_UNDEFINED = -1;
+
     static final byte NOTIFICATION_EMAIL = 0;
     static final byte NOTIFICATION_SMS = 1;
     static final byte NOTIFICATION_TWITTER = 2;
     static final byte NOTIFICATION_FACEBOOK = 3;
+
+    static final byte ICON_NOTIFICATION_GENERIC = 1;
+    static final byte ICON_TIMELINE_MISSED_CALL = 2;
+    static final byte ICON_NOTIFICATION_REMINDER = 3;
+    static final byte ICON_NOTIFICATION_FLAG = 4;
+    static final byte ICON_NOTIFICATION_WHATSAPP = 5;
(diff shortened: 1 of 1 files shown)


roidelapluie:
Travis is failing because master branch is failing

ashimokawa:
Thank you for the PR, however I have two comments:
1) The icons IDs (even more of them) are  in  the Pebble SDK Pebble/common/phonesim/timeline/layouts.json, so they are not a secret and known. Also we should not prefix them with NOTIFICATION_ because they serve different purposes also.

2) You seem to have made up a new constant named NOTIFICATION_GENERIC = 4 (or is this official?), which could break Endpoint 3000 (for firmwares <3.x) notifications because the constant will be directly encoded in the protocol and is invalid AFAIK. 
I understand that you just wanted to distinguish SMS and generic notifications in the 3.x notification encoder. I will find another way, maybe an an enum to clearly seperate it from the constants which are actually important for the encoded protocol.

roidelapluie:
@ashimokawa thanks! Care to review this one?

roidelapluie:
1. I have taken the list from the SDK
2. Solved by always sending the same type of notification for endpoints notifications

roidelapluie:
wait it is never email

roidelapluie:
if (mForceProtocol || type != NOTIFICATION_EMAIL) {

this means that we use endpoint notifications only when type = NOTIFICATION_EMAIL, am I right?

ashimokawa:
correct.
The reason is that I am unable to set the icon on Endpoint 3010 (2.x) notfications, so I use Endpoint 3000 (1.x) notifications for email.

Please see my comment above.

roidelapluie:
Thanks, I have added a NOTIFICATION_UNDEFINED byte to replace the old one

roidelapluie:
@ashimokawa I have changed my comment to just 'type is always NOTIFICATION_EMAIL'.

roidelapluie:
Actually that type is never sent to the watch for 2.x and 3.x protocols, in 2.x it is not there and in 3.x we convert that to icon_id.

ashimokawa:
Yes, I know, this is only used for 1.x notification and unfortunately I misused the constants for the type parameter, hence the confusion.
The reason why it is not used for 2.x notification is that I am unable to set the icon id there (attribute id 4  should be correct but results in crashes when looking at the list of notifications on the Pebble)

roidelapluie:
Do you expect something more from me on this pull request? I think we could safely look to set different colors for each NOTIFICATION_TYPE but if I figure out how to do this it will be in a second pull request.

ashimokawa:
I will merge this today.
For the colors, I think I know how this works.

roidelapluie:
Thanks for your feedback and many thanks to all the team to allow us to free our gadgets!

PR #122, opened 2015-09-13
Pebble: Support Colors in notifications

roidelapluie:


ashimokawa:
Maybe we should put these in a static class called PebbleColors?
That might also make sense for the Icon IDs.

roidelapluie:
I have limited java knowledge but I will do it

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java
```diff
@@ -744,8 +820,8 @@
             actions_length = (short) (ACTION_LENGTH_MIN * actions_count + dismiss_string.length() + open_string.length());
         }
 
-        byte attributes_count = 1; // icon
-        short attributes_length = (short) (7 + actions_length); // icon
+        byte attributes_count = 2; // icon
```
roidelapluie: @ashimokawa care to review this line? not sure about it but it works.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java
```diff
@@ -744,8 +820,8 @@
...
+        short attributes_length = (short) (14 + actions_length);
```
roidelapluie: @ashimokawa care to review this line? not sure about it but it works.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java
```diff
@@ -698,30 +766,38 @@
 
         String[] parts = {title, subtitle, body};
 
-        int icon_id;
+        byte icon_id;
```
roidelapluie: @ashimokawa Here I switched to byte

ashimokawa:
Your code in incorrect, you specify 1 as length (for byte) but the put 4 bytes (int), You should  correct it by:
1) not adding 7 to the length but 4
use put() instead of putInt() for the color

Your coded produces invalid protocol, the color is picked up anyway because it is little endian and LSB comes fist.

ashimokawa:
icon id is int, color is byte, please fix, also in the constants.

roidelapluie:
color in byte does not work

roidelapluie:
why should we have color_id in int at this point? we convert it later in int anyway (at the putInt).

ashimokawa:
Correct, but it is confusing to have it as byte even though it is always an int in the protocol.

ashimokawa:
your code is still wrong as you specify 1 as length and then put 4 bytes in the prococol.
And java sucks, so you have to cast 0b11111111 to byte, then it works.

roidelapluie:
short attributes_length = (short) (11 + actions_length);

is not working

roidelapluie:
ok, I will test that

ashimokawa:
it works here, you also have to replace putInt() with put()

roidelapluie:
I am getting closer:

```
/home/roidelapluie/dev/Gadgetbridge/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleProtocol.java:814: error: no suitable method
 found for put(int)
        buf.put(color_id);
           ^
    method ByteBuffer.put(byte) is not applicable
      (argument mismatch; possible lossy conversion from int to byte)
    method ByteBuffer.put(byte[]) is not applicable
      (argument mismatch; int cannot be converted to byte[])
    method ByteBuffer.put(ByteBuffer) is not applicable
      (argument mismatch; int cannot be converted to ByteBuffer)
```

ashimokawa:
please change
        buf.putShort((short) 4); // length of int
        buf.put(color_id);
to
        buf.putShort((short) 1); // length of int
        buf.put(color_id);

and also make color_id byte

roidelapluie:
here you go, testing on my watch

roidelapluie:
very nice, it works.

ashimokawa:
Thanks! ;)

PR #124, opened 2015-09-13
Pebble: Cleanup of Code and Colors variables

roidelapluie:
* Rename COLOR_ prefix from color names because they are in an explicit class
* Move ICONS_ variables to their own static class


ashimokawa:
sorry, I did the same at the same time

roidelapluie:
NP, thanks :)

ashimokawa:
Well, I also made icon ids int and chose different names.
Also I am not sure if the files should be in that package or in the same package as PebbleProtocol

PR #125, opened 2015-09-13
Start working on quick replies

roidelapluie:


roidelapluie:
Does that support only one level answers? otherwise we could do:

Reply > +0436438294 > [quick, replies]

That would be an easy way to get the phone number

ashimokawa:
We do not get any detail back from the pebble just the action id and the chosen reply id.

roidelapluie:
this is too complex for my current knowledge, closing

ashimokawa:
Thank you anyway, nothing in the PR was bad, it was just missing something.

Plus the real challenge is to find a _smart_ way of mapping a message id back to a phone number and a _good_ place in the code to send out the SMS.

roidelapluie:
@ashimokawa do you plan to add this to your roadmap?

ashimokawa:
It is on the roadmap in my head, yes ;)

PR #140, opened 2015-10-07
Add miband bluetooth codes for fw 5.15.8.21

platinumthinker:
I sniff my mi band (v.1.1 with white LED) and add new codes.


platinumthinker:
PS: I use BLExplorer from my phone.

cpfeiffer:
Thanks -- I don't want to merge this as is, as I'd like to mark the additional characteristics as for this fw only. If you add the comments yourself, I'll merge it, otherwise I'll do it later.

platinumthinker:
I am in the process of further details

danielegobbetti:
1802 and 2a06 are already in because of the new firmware for the old miband. They are part of the standard bt sig designed uuids and are documented in the corresponding classes. So they should be removed from here as well.
--

platinumthinker:
thanks for tips

cpfeiffer:
@platinumthinker Do you want to update this PR or shall we do it ourselves?

PR #144, opened 2015-10-18
Initial support for newer MI devices.

sarg:
Hi, as an owner of MI1A (in yesterdays comment I mistyped it as MI1S), I am willing to add support for it. So, here is my first baby step - try to pair :)

I also added notification toast when no device support is found. But it will be better to close PairingActivity and show error message.

 After changes device still not pair but now there is an communication error. I have found that MI1A UserInfo differs from MI.


cpfeiffer:
That was quick, thanks a lot :-) Didn't even know that there's yet another place with the hardcoded "MI" name.

PR #146, opened 2015-10-19
Fix DeviceInfo checksum function.

sarg:
Hi, fixed bug in checksum.
Found it when my second MI1A didn't wanted to pair :)


cpfeiffer:
Thanks!

PR #235, opened 2016-02-28
Use external browser for configuring pebble apps

danielegobbetti:
Needed steps before this code can be merged:
- [x] Save the app keys to map the configuration values from the webpage to pebble keys.
- [x] Save the configuration javascript contained in the pbw file.
- [x] Create a local webview with javascript access that is able to emulate the javascript Pebble object (needed by the original javascript configuration files).
- [x] Interact with the connected device to get the real platform (aplite, basalt, chalk, ...)
- [x] Override the emulated Pebble js methods to use an external browser.
- [x] Convert the JSON returned by the website to local keys.
- [x] Handle booleans correctly (int32?)
- [x] Interact with the connected device to push the message to the watch app.
- [ ] fix 64k limit for .js files
- [x] Support further Pebble js methods like Pebble.getWatchToken().hashCode();  found on squared4 watchface
- [ ] Style the local webpage.
- [ ] Properly warn the user if there is no configuration option (the app does not contain js file, or we fail to support it in some way)
- [ ] Consider / add support for https://github.com/pebble/clay/

Limitations:
- The pebble configuration site must support the return_to parameter.
- inline configuration pages (data url) support has yet to be investigated

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -53,6 +53,7 @@
             android:label="@string/preferences_miband_settings"
             android:parentActivityName=".activities.SettingsActivity" />
         <activity
+            android:launchMode="singleTop"
             android:name=".activities.AppManagerActivity"
             android:label="@string/title_activity_appmanager"
             android:parentActivityName=".activities.ControlCenter" />
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -257,6 +257,22 @@
                 android:name="android.appwidget.provider"
                 android:resource="@xml/sleep_alarm_widget_info" />
         </receiver>
+
+        <activity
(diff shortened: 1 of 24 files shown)


ashimokawa:
I don't mind if we merge this early.
What does clay support mean for us?

ashimokawa:
This branch makes it impossible to install morpheuz.
The .js file seems to be > 64k

danielegobbetti:
I guess you answered to yourself re what clay could mean to us: bigger local JS. Possibly multiple js files as well...

ashimokawa:
Merged early, sorry ;)

danielegobbetti:
Ok this has still to be tracked because of checkboxes, should I create a new issue?

ashimokawa:
okay, go ahead - seems we have some work to to ;)

danielegobbetti:
remaining work moved to #251

PR #247, opened 2016-03-10
Add button to allow selecting files to install

madbrenner:
This pull request adds a button to AppManagerActivity to allow users to select a file to install to their device.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/AppManagerActivity.java
```diff
@@ -41,6 +42,7 @@
     public static final String ACTION_REFRESH_APPLIST
             = "nodomain.freeyourgadget.gadgetbridge.appmanager.action.refresh_applist";
     private static final Logger LOG = LoggerFactory.getLogger(AppManagerActivity.class);
+    private static final int READ_REQUEST_CODE = 42;
```
danielegobbetti: That's the answer, I guess :-)

madbrenner:
I thought I would close this pull request as it appeared to have been added to the repo, but that seems to have reverted the changes, so I am reopening.
I apologize for the confusion.

ashimokawa:
An icon inside the title bar (or however this is called would be better I guess), but anyway I will think about merging this as-is.

madbrenner:
Just to clarify, do you mean the overflow menu, like how in the Control Center activity you can open the menu for options such as "Connect new Device"?
I imagine that shouldn't be difficult, I will look into making that change.

ashimokawa:
I dont know how it is called ;)
I thought about some "+" icon right inside the action bar replacing the current button.
If you want to do it - fine. If not, tell me, then I will.

madbrenner:
I would like to make the change myself, but it likely won't get done until this weekend, in case that is an issue.

ashimokawa:
@madbrenner 
No problem, take your time.

madbrenner:
Sorry for taking so long with this, but I removed the button and made a menu entry for it.
Let me know if that looks good, or if I should change anything else.

ashimokawa:
Thanks.
I see a small problem.
onResume() will always be called after onCreate() so in case the activity is started first, the list will refreshed twice. 

Also on Firmware 2.x, where we really can get lists of installed apps by asking the watch, the list will be completely empty after returning installation.

I can merge this and fix the problems myself. Will do that right after the 0.9.1 release

ashimokawa:
Sorry for being that late. I did something very similar now, with a FAB and a bit simpler, see the referenced commit.

Anyway thanks for the suggesting and your contribution.

Issue #232, opened 2016-02-24
Inaccurate sleep/wake recognition


official-jkim:
I'm not sure if Gadgetbridge processes raw data from Mi Band or just simply fetches it, but Mi Band doesn't recognise wakeup properly, it shows that my activity is 'light sleep' not 'activity'. There are sporadic short 'activity' graphs, but no sign of fully 'waking up'. This inaccurate sleep detection is resulted in much longer sleep hour, such as 12 hours of sleep instead of 8 hours of actual sleep.

or maybe it's because my Mi Band is broken.

danielegobbetti:
@official-jkim we merely collect the raw data (minute samples) from the device and put them on the graphs.

We started collecting ideas / algorithms for interpreting the data in the following wiki page:
https://github.com/Freeyourgadget/Gadgetbridge/wiki/Activity-analysis

Any further ideas and contribution is much welcome.

See also #135 and other issues...

Samarium151:
All of these devices are not particularly accurate for measurements of sleep.  One big issue with them is the manufactures mostly implicitly over sell their accuracy.   (heck rigorous checks of the pedometer functions also do  not pan out well either. )  I added one reference to the wiki page above.  Gadgetbridge documentation should in my opinion have some explanation of this.

danielegobbetti:
I can happily provide more details on how raw* data is coming from the different devices (original miband and pebble health). That would be a description of how things are, not an explanation on how activity tracking works.

Thanks for contributing to our wiki! Do you think that what I proposed above could help you giving the explanation you have in mind?

*We usually refer to this as "raw" data but in fact they are already manipulated on the devices. For both miband and pebble we get per-minute aggregated data, for instance.

official-jkim:
I think my Mi Band is broken, as you said even Mi Band's raw data is manipulated on the device. Now it doesn't recognise 'wakeup' at all, recognises any activity as 'sleep' constantly. After even most intensive shake movement and highest activity value, Mi Band recognises it as very short 'activity' and then go back to recognising it as 'light sleep'. Oh well, time to throw it away.

Samarium151:
I think that's a good start.  I'll try to find some more objective stuff on the accuracy to reference.  The manufacturers have no incentive to publish data that says there devices arent' accurate so it may be har dto come by.  I'd be interested to see what data the devices are putting out as that would help in trying to figure out what if any analysis gadgetbridge should do with it.

[...]

PR #273, opened 2016-04-03
Heartrate sleep support Mi1S

computerlyrik:
Add a setting to preferences to enable support HearRate for Sleep detection.

Had some problems with the Preferences Framework due to using a Boolean Checkbox.

Settings are not synced on leaving the settings screen, only on reconnecting the device?

References https://github.com/Freeyourgadget/Gadgetbridge/issues/232


cpfeiffer:
I'll make some changes to the initial enabling of the sensor in order to fix a bogus "Heart rate not supported" message that I get when initiall connecting. Then I'll check the results over night. Let's see how it goes :-)

cpfeiffer:
I merged this PR already, because it looks pretty good. Looking forward to the results tomorrow.

cpfeiffer:
Oh, and thanks! :+1:

cpfeiffer:
I did get heart rate values from the night, so it appeared to work :-)

There are some bogus values (0 and 255) that we need to filter out somewhere, but that's expected.

Issue #89, opened 2015-07-29
Pebble: Better automatic reconnect


derlaft:
Official pebble app seeks for bt device connections and re-establishes a connecting to watch once it's connected.

This functionality is missed in gadgetbridge, which forces users to reconnect manually.

ashimokawa:
@derlaft 
Gadgetbridge tries to reconnect but only for a limited amount of time.
I don't know how the official App behaves, but I noticed that the Pebble itself tries to connect to the Phone which we currently do not support. Maybe Pebble uses that passive re-connection approach in their App?

I could add that feature and make it an option. Something like "listen to incoming connections from the Pebble". I have no idea however under which conditions and when the pebble connects and to which Phone if multiple Phones are paired.

roidelapluie:
I would also like an 'agressive reconnects' option. gadgetbridge would never stop trying reconnecting to the watch when enabled.

glitchphan:
Just to throw some more weight on this issue, I have to manually reconnect the app several times a day. There are plenty of times when I will get far away from the phone, and it automatically reconnects when I get within range again, but sometimes it'll disconnect when I'm just a few feet away and never reconnect.

ashimokawa:
I will add either an option or raise the number of reconnect attempts.
@glitchphan 
Your issue with the disconnects withing a few feets range is strange. The only thing I could think of is that the service got killed and restarted (out of memory). In any case we need a solution for that scenario.

glitchphan:
This likely had nothing to do with BT range, and I'm betting I need to turn on some kind of logging to really know what's going on. I was in the next room over from the phone and the Pebble Time vibrated like it does when it disconnects. When I came into the room with the phone, I waited a little bit but it never connected. Turning on the screen and pulling down the notification drawer showed Gadgetbridge, but underneath that it just said "Gadgetbridge is running" rather than "... is initialized" or "... is not connected", the way it usually does. I have not noticed this behavior before when it doesn't reconnect on its own, but this is the new version. As for memory, I have 2gb, and a cat /proc/meminfo shows 500mb cached, 30mb in buffers, and 70mb free. Never really paid attention to these numbers before so I don't know how normal it is, but there is obviously more running/active at the moment as I type this than there was while it was sitting on my nightstand.

ashimokawa:
You can turn on logging in Gadgetbridge's settings. Maybe then we get an idea of what happens.
I don't think this has been introduced lately.

[...]

PR #276, opened 2016-04-06
Pebble: Smarter reconnection attempts

roidelapluie:
Sleep several seconds between reconnection attempts:
One second after first attempt, two after the second, and so on.

refs #89


ashimokawa:
I am pretty sure I had a reason NOT to sleep on that thread.
But since I can't remember I will merge this and see how it works.

ashimokawa:
Ok now I remember
ANR when quitting or disconnecting during the sleep

(this is because we try to join the thread)

roidelapluie:
Can we sleep X times 1 second to avoid this, instead of one time X seconds?

ashimokawa:
No that would  not matter.
I might interrupt the thread instead of joining.

PR #278, opened 2016-04-09
UI Changes

bytehala:
- FAB for adding devices, removed the action from overflow
- Added background logo for empty ControlCenter
- Change the Activities to AppCompat


bytehala:
Feel free to squash.
https://github.com/blog/2141-squash-your-commits

bytehala:
Feel free to squash.
https://github.com/blog/2141-squash-your-commits

cpfeiffer:
Aweseome! I'll have a look at this in the next days because of other duties this weekend. Thanks!

ashimokawa:
@bytehala 
Cool, thanks!

ashimokawa:
@bytehala 
Can you remove 
"Increase JVM heap size" (fc72c56)
from the PR?

ashimokawa:
@bytehala 
I only had a quick look but so far I like what I saw ;)

Debug activity crashes because of 

getActionBar().setDisplayHomeAsUpEnabled(true);

which can be removed after your changes AFAIK.

Also the Blue Icon shadow looks "cut off" in an ugly way. Maybe only on my phone, but I think this can be fixed quite easily.

Thanks again

bytehala:
Thanks, I'm taking a look right now and it looks like there are other places it could crash, could you make a quick check?:

- AppBlacklistActivity
- AppManagerActivity
- ConfigureAlarms
- DebugActivity
- ExternalPebbleJSActivity
- FwAppInstallerActivity

Regarding the JVM heap size, travis doesn't like it if I remove it but sure I will set it to what you have at 256m.

For the cut-off button, I don't know what to do yet, as it's bundled in the design library. I'm thinking add a pre-Lollipop (or some-dpi modifier?) version of the layout which doesn't use Google's library.

Let me know what you think. Thanks.

ashimokawa:
The activities all crash since getActionBar() returns null.

For our take on fixing jdk7 build see:

059c7d0b

(We didn't set the heap to 256mb, look closer ;) )

The cut-off icon shadow might be due to our padding. It is not visible on my galaxy tab but on my s4 mini. Should be easy to fix.

bytehala:
It is done.

ashimokawa:
@bytehala 

Thanks!

Is there any reason for keeping setDisplayHomeAsUpEnabled(true)?
AFAIK I had to had that because we did not have AppCompat - but since you changed that it should not be necessary. Also getActionBar() should return always null now (at least I tested it on Android 4.4 and 6.0)
 
Also another minor problem is that when connected to a pebble, on small screens the floating button is above the text that tells you to tap on the connected device to open the App Manager, making the last part of the string unreadable.

bytehala:
Okay, I removed those and it looks the same, so no harm done.

Regarding the hint text, I changed it to sit above the FAB, so that they don't overlap but have you considered removing it? I mean, it's already obvious to the user that they have to tap the connected device or maybe I'm missing something. Just a thought. Anyway it looks like this now:

Preview

bytehala:
I changed the ProgressBar and added grayscale.

Device New UI

cpfeiffer:
Great work, thanks a lot! I've squashed two commits and pushed them to master.

We have more plans to change the UI (e.g. the context menu sucks), but again, after the db changes. I'll add some more content to https://github.com/Freeyourgadget/Gadgetbridge/wiki/Brainstorming-new-UI

Feel free to chime in whereever you want :-)

PR #281, opened 2016-04-09
Entity generation fix, gradle and travis build script fixes

bytehala:
It's now able to generate the entities onto the `app` module, and the mainClassName script error has been solved.

I think that the entities package should be put in `.gitignore` but I will defer that decision to @cpfeiffer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: android
+
 jdk:
   - oraclejdk8
   - oraclejdk7
+env:
+  - GRADLE_OPTS="-Xmx2048m -XX:MaxPermSize=512m"
+
 android:
   components:
     # Uncomment the lines below if you want to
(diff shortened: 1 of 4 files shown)


cpfeiffer:
Thanks, I've cherrypicked 56dde47, I hope that's ok for you?

Regarding the `gen/` folder -- I wanted to generate the entities into a separate folder than `src/main/java`, but gradle wouldn't compile them in `src/main/gen`, as I had hoped. Well, no need to be perfect. I ignored *.java in the entities package, so that at least the package is visible by default.

bytehala:
Yeah, I was thinking the entities could be in their own module, since they have no dependency on the others so then you have `app`, `gen-entities`, and `GBDaoGenerator`. This architecture is usually encouraged, as they don't have Android dependencies... then you can test the pure Java modules with JUnit. I thought I'd suggest it first, since you're the one mainly working on it. What do you think?

PR #306, opened 2016-05-19
Detects if Do Not Disturb is in use

normano64:
I felt that Do Not Disturb was missing when using my Mi Band 1S.
I'm not used to develop for Android or with java but everything seems to work as expected.

Added a preference to make notifications comply with the Do Not Disturb mode.
Can handle sms and phone calls from priority senders when in Priority only mode, but not events or reminders.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
@@ -1,15 +1,20 @@
 package nodomain.freeyourgadget.gadgetbridge;
 
 import android.app.Application;
+import android.app.NotificationManager.Policy;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.res.Resources;
+import android.database.Cursor;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Build.VERSION;
 import android.preference.PreferenceManager;
+import android.provider.ContactsContract;
+import android.provider.ContactsContract.PhoneLookup;
 import android.support.v4.content.LocalBroadcastManager;
 import android.util.Log;
 import android.util.TypedValue;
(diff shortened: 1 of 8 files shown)


cpfeiffer:
Cool, will have a look at it in the evening, thanks!

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
```diff
@@ -247,6 +252,41 @@ public static boolean isRunningLollipopOrLater() {
...
+                exists = true;
+                starred = cursor.getInt(cursor.getColumnIndexOrThrow(PhoneLookup.STARRED));
+            }
+        } finally {
+            if (cursor != null) {
+                cursor.close();
+            }
+        }
+        switch (prioritySenders) {
+            case Policy.PRIORITY_SENDERS_ANY:
```
cpfeiffer: This case could be moved to the beginning of the method, so that the lookup is skipped altogether.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/SettingsActivity.java
```diff
@@ -107,6 +110,12 @@ public boolean onPreferenceChange(Preference preference, Object newVal) {
 
         });
 
+        if (!GBApplication.isRunningMarshmallowOrLater()) {
+            pref = findPreference("notification_filter");
+            PreferenceCategory category = (PreferenceCategory) findPreference("pref_key_notifications");
+            category.removePreference(pref);
```
cpfeiffer: Nice :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/K9Receiver.java
```diff
@@ -33,6 +34,17 @@ public void onReceive(Context context, Intent intent) {
                 return;
             }
         }
+        if (prefs.getBoolean("notification_filter", false) && GBApplication.isRunningMarshmallowOrLater()) {
```
cpfeiffer: The following code looks partly duplicated among the different receivers. Can you centralize this somewhere? Either in a base class or a helper method?

For example, you could add a method getGrantedInterruptionFilter() that would return the actual filter if all conditions are met, and INTERRUPTION_FILTER_ALL for any failed condition.

cpfeiffer:
Very nice job, thank you very much. Could you have a look at my comments? :-)

normano64:
Thank you. I will try to fix it.

normano64:
I think this is better, based on your feedback.
I also realized that I forgot to check if the Priority Categories was on, fixed now.

Thanks for you work on the app.

cpfeiffer:
Excellent, thank you very much! :+1

PR #315, opened 2016-05-24
MusicPlayBackReceiver set track, artist and album to the artist membe…

AnotherStranger:
…r of MusicSpec.

Now the assignment of artist, track and album is correct


cpfeiffer:
Good catch, thanks :-)

PR #321, opened 2016-06-03
Add support for shifting the device time by N hours

pragmaware:
Mi-band firmware seems to gather sleep data only if sleep is initiated in a certain period of the day, probably between 10pm and some hour in the middle of the night. This means that shift workers, or just anyone that has the habit to go to sleep at a very different time can't use mi-band for sleep tracking.

This commit attempts to work around this limitation. It adds an option that specifies a device time offset in hours. All the times sent to the device are shifted by this offset and all the times received from the device are shifted "back". In this way mi-band thinks to be in a different time zone and gathers sleep data at different hours but data view on the app side is unaffected.

Example of usage: if you usually go to sleep at 6am, set the offset to -8 hours. At 6 am the device will think it's still 10pm and will begin gathering sleep data.


cpfeiffer:
I haven't checked your patch yet, but your idea is excellent, I love it!

cpfeiffer:
Great job, thanks a lot!

pragmaware:
Great! Thanks for merging :)

PR #325, opened 2016-06-07
Set music info for PocketCasts

stepardo:
PocketCasts tells about its current media state via notifications. This
patch tries to parse incoming notifications from PocketCasts and if
successful tells the device about it. Currently supported are track and
artist.

I tested this with my (original) pebble. I do not know if this can be enhanced with track lenght and current progress as I was not able to squeeze this info out of the notification. Also, the device setMusicInfo would have to be improved to set the current progress. Both should be possible as the pebble android app apparently can do that.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
@@ -218,7 +218,7 @@ public class DebugActivity extends GBActivity {
                 musicSpec.artist = editContent.getText().toString() + "(artist)";
                 musicSpec.album = editContent.getText().toString() + "(album)";
                 musicSpec.track = editContent.getText().toString() + "(track)";
-                musicSpec.duration = 10;
+                musicSpec.duration = 10 * 1000;
                 musicSpec.trackCount = 5;
                 musicSpec.trackNr = 2;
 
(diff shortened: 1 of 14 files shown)


stepardo:
After some fiddling with Android programming, I have a solution to "the full feature set" (tm) ;-)

That means: setting the duration, current position, playback rate, playback state etc pp. It might even be a generic solution, not only for pocketcasts, but I will have to do more testing.

What I am trying to say is: Please do not yet merge this changeset because I am working on a much better and more featurefull solution. Thanks.

cpfeiffer:
Cool, looking forward to it :-)

ashimokawa:
I didn't know this was possible. Looking forward to some magic ;)

stepardo:
So there it is. It is working very well on my setup (nexus phone, android 5.0 and original pebble). The caveat is that it requires the device to run android 5.0 or newer.

ashimokawa:
Thanks! This looks great. I will try it with some players sometime today.

A few remarks after a first glance:

- You separated music info and music state, I am not sure if this is good. Sure from a pebble point of view this is what the pebble protocol does. But since Gadgetbridge is at least meant to offer some kind of framework to support multiple devices this is maybe a bit to pebble centric. We can leave it like this for now, but I might merge these later ;)
- For track length milliseconds are not good for a public API I think, seconds are good enough, the last commit should be reverted
- we should not use magic values like 0x01, there should be constants like in MusicSpec.java

ashimokawa:
I tested the code at it is working very well with the CM Music app :+1: 
Shall I just merge your branch and then address some minor points myself (like constants and milliseconds)?

stepardo:
Hi,

thanks for testing it. Glad it worked for you!

I addressed two of your complaints.
1. I reverted the milliseconds change. This is a matter of taste, and I let you decide;-)
2. I introduced names for the different play states. This was bothering me as well.

As for the third one, it might be useful to have musicinfo and state separated. For the mediasession notifications it obviously does not need to be split as this contains both information, and we generate two intents when really we only need one. However, there is this "com.android.music.metachanged" broadcast intent, which only affects the music information but not the state. In that situation we must not update the state as we (currently) don't store it anywhere. Resetting the state would also break the pebble music app. I suspect there are other sources of information that I am currently not aware of that also split this information. It makes sense to split it as the media information needs to be updated only when a new media file is being played, but the state needs to be updated whenever the user pauses/continues playback.

Thanks again for your great work!

Best,
Steffen

ashimokawa:
Thank YOU for your contribution :+1:

PR #386, opened 2016-09-12
Added miband2 connection support

JohnnySun:
Added miband2 connection support, now when you bound miband2, it need to tap to confirm. and we also do some key exchange to auth miband2, after auth, we can keep miband2 connection and get other gatt Characteristic value.

its on doing , now i just added the key exchange and tap confirm with band2, after connect, you can use nodic to get other Characteristic value


JohnnySun:
screenshot_20160913-013720
Here's after connect use nordic to notification heartt rate measurment

cpfeiffer:
Awesome, very cool! I hope Mi2 support will finally get some progress :-)

PR #387, opened 2016-09-13
fix somebug and rewirte the auth process

JohnnySun:
this time i fixed the bug on auth and change auth peogress. but i also need your help.
as i know,  at our app first connect miband2, we need send { 0x01, 0x08, (byte) 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40, 0x41, 0x42, 0x43, 0x44, 0x45 } to UUID_UNKNOWN_CHARACTERISTIC9 at the auth key.
but i am not found what func called when the first time device discovery.

so can you tell me whats func call on the deivce first init.


JohnnySun:
oh theres another question.
i commit the .requestDeviceInfo(builder) and .requestBatteryInfo(builder); because if i wont do this, i cant do my mBluetoothGatt.writeCharacteristic(temp) on my boardcastreceiver

PR #390, opened 2016-09-13
Finally, miband2 connect complete.

JohnnySun:
This time i fixed the bug on auth and change auth peogress. but i also need your help.
As i know, at our app first connect miband2, we need send { 0x01, 0x08, (byte) 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40, 0x41, 0x42, 0x43, 0x44, 0x45 } to UUID_UNKNOWN_CHARACTERISTIC9 at the auth key.
but i am not found what func called when the first time device discovery.

So, can you tell me whats func call on the deivce first init.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
@@ -112,13 +112,7 @@ public class MiBand2Support extends AbstractBTLEDeviceSupport {
                 byte[] response = intent.getExtras().getByteArray(DeviceService.EXTRA_MIBAND2_AUTH_BYTE);
                 BluetoothGattCharacteristic temp = getCharacteristic(MiBand2Service.UUID_UNKNOWN_CHARACTERISTIC9);
                 temp.setValue(response);
-                if ((temp.getProperties() & (BluetoothGattCharacteristic.PROPERTY_WRITE | BluetoothGattCharacteristic.PROPERTY_WRITE_NO_RESPONSE)) != 0) {
-                    Log.d("HELLO", "its looks ok");
-                }
-                if (!mBluetoothGatt.writeCharacteristic(temp)) {
-                    Log.d("ERROR", "ERROR");
-                }
-                Log.d("HELLO", "HELLO");
+                mBluetoothGatt.writeCharacteristic(temp);
 
             }
         }
@@ -173,7 +167,8 @@ public class MiBand2Support extends AbstractBTLEDeviceSupport {
 // this is apparently not needed anymore, and actually causes problems when bonding is not used/does not work
 // so we simply not use the UUID_PAIR characteristic.
 //                .pair(builder)
-                .testInit(builder);
+                .testInit(builder)
+                .setInitialized(builder);
                 //.requestDeviceInfo(builder)
                 //.requestBatteryInfo(builder);
 //                .sendUserInfo(builder)
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
@@ -7,6 +7,9 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.net.Uri;
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Message;
 import android.support.v4.content.LocalBroadcastManager;
 import android.util.Log;
 import android.widget.Toast;
@@ -55,6 +58,7 @@ import nodomain.freeyourgadget.gadgetbridge.service.btle.BLETypeConversions;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.BtLEAction;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.GattCharacteristic;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.GattService;
+import nodomain.freeyourgadget.gadgetbridge.service.btle.Transaction;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.TransactionBuilder;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.actions.AbortTransactionAction;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.actions.ConditionalWriteAction;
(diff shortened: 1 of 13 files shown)


JohnnySun:
I think i need use builder.write to replace BluetoothGatt.writeCharacteristic();
Next step is to send Auth key ' 0x01, 0x08, (byte) 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40, 0x41, 0x42, 0x43, 0x44, 0x45 ' when device first connect Gadgetbridge.
---
BTW, we need to clean up the Miband2Servie.java, there are too many unused code and lint warning XD.

gradle.properties
```diff
@@ -1,19 +1,19 @@
-# Project-wide Gradle settings.
-
-# IDE (e.g. Android Studio) users:
-# Gradle settings configured through the IDE *will override*
-# any settings specified in this file.
-
+## Project-wide Gradle settings.
```
cpfeiffer: What are the reasons for the changes to this file? Did you copy it from somewhere or what do you want to achieve with tme?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
```diff
@@ -104,12 +108,25 @@ public void onReceive(Context context, Intent intent) {
 
             } else if (s.equals(BatteryInfoProfile.ACTION_BATTERY_INFO)) {
                 handleBatteryInfo((nodomain.freeyourgadget.gadgetbridge.service.btle.profiles.battery.BatteryInfo) intent.getParcelableExtra(BatteryInfoProfile.EXTRA_BATTERY_INFO));
+            } else if (s.equals(DeviceService.ACTION_MIBAND2_AUTH)) {
+                byte[] response = intent.getExtras().getByteArray(DeviceService.EXTRA_MIBAND2_AUTH_BYTE);
+                BluetoothGattCharacteristic temp = getCharacteristic(MiBand2Service.UUID_UNKNOWN_CHARACTERISTIC9);
+                temp.setValue(response);
+                if ((temp.getProperties() & (BluetoothGattCharacteristic.PROPERTY_WRITE | BluetoothGattCharacteristic.PROPERTY_WRITE_NO_RESPONSE)) != 0) {
+                    Log.d("HELLO", "its looks ok");
```
cpfeiffer: Can you please use the static slf4j LOG variables instead of android.util.Log directly?

We use slf4j to log both to android.util.Log and also to a file that can be inspected later (logging to file is configurable).

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
```diff
@@ -958,6 +980,10 @@ public boolean onCharacteristicWrite(BluetoothGatt gatt,
         } else if (MiBandService.UUID_CHARACTERISTIC_CONTROL_POINT.equals(characteristicUUID)) {
             handleControlPointResult(characteristic.getValue(), status);
             return true;
+        } else if (MiBand2Service.UUID_UNKNOWN_CHARACTERISTIC9.equals(characteristicUUID)) {
+            LOG.info("KEY AES SEND");
```
cpfeiffer: Yes, this is the correct LOG :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
```diff
@@ -104,12 +108,25 @@ public void onReceive(Context context, Intent intent) {
...
+                }
+                Log.d("HELLO", "HELLO");
+
             }
         }
     };
 
     private volatile boolean telephoneRinging;
     private volatile boolean isLocatingDevice;
+    private BluetoothGatt mBluetoothGatt;
```
cpfeiffer: Is it necessary that you remember the BluetoothGatt here and directly write to it? I'd prefer if you could use the TransactionBuilder /BtleQueue to do this.

If this doesn't work for some reason, I'd like to figure out why it doesn't work and if we could improve the infrastructure to make it work. 

But in the meantime, feel free to keep it that way so that you can progress without waiting for better infrastructure.

cpfeiffer:
First of all: great job, very nice!

Now to the feedback :-)

> So, can you tell me whats func call on the deivce first init.

At the moment we do not have a "first time init function", because we didn't need something like that yet. The closest we have right now is the DeviceSupport.pair() method which is only called when connecting to a new device using the discovery activity. Would that be sufficient or do you really need to track if we have ever sent the auth key or not?

> I think i need use builder.write to replace BluetoothGatt.writeCharacteristic();

Yes, using builder.write and send everything through the queue is the preferred way to do it. If there is a problem doing it that way, please tell, maybe we need some more infrastructure for such kind of ping-pong communication between the devices.

> BTW, we need to clean up the Miband2Servie.java, there are too many unused code and lint warning XD.

Yeah, initially we thought that Mi Band 2 would be similar to Mi Band 1, so we took that as starting point.

Last: I'll merge this once I know if the changes to `gradle.properties`  are really intended for all of us.

Thanks!

JohnnySun:
oh, the grade.properties looks like the Android studio optimize automatically when I optimize them... I will rollback it and make a new pull request later time. Sorry for that mistake😂

JohnnySun:
The reason of change is I set grade proxy... Because Chinese fxxxing great firewall😂

JohnnySun:
Closed #390

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
JohnnySun: I 'll change with  the static slf4j LOG at next commit. :)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband/MiBand2Support.java
JohnnySun: But, I still have a question, thats whats the way to keep a builder on this class.  its looks like hold a private reference is not a good way.  
Maybe use `TransactionBuilder performInitialized(String taskName)` is batter than keep priference reference ?

cpfeiffer: Just a quick answer before leaving home...yes, typically.we use TransactionBuilder to perform a series of read/write actions.

TransactionBuilders are short lived and cannot be used anymore after queueing their result. But you can also use the BtLEQueue directly, e.g. getQueue().add(new WriteAction(...))

Might be that we need to add a insert() or addFront() method to BtLEQueue so that it will execute the given action right away instead of adding it to the end of the list.

Almtesh:
Hi,
I installed the 0.12.2 version from F-droid and the connection does not succed. Is it in the next version?

PR #391, opened 2016-09-13
Rollback gradle.properties

JohnnySun:
For more information, see #390

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/database/DBHandler.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/database/DBHandler.java
@@ -3,7 +3,6 @@ package nodomain.freeyourgadget.gadgetbridge.database;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
 
-import nodomain.freeyourgadget.gadgetbridge.GBApplication;
 import nodomain.freeyourgadget.gadgetbridge.entities.DaoMaster;
 import nodomain.freeyourgadget.gadgetbridge.entities.DaoSession;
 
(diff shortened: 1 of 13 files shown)


ashimokawa:
@JohnnySun 
Thanks for your contribution! I am sure @cpfeiffer will have a look soon.

Please note you do not have to close and open a new Pull Request every time.
It is enough just to push to your branch and the new commits will add to the existing PR.
That makes commenting much easier.

JohnnySun:
ok, Thanks for ur remind😀， i will back  to do these work after chinese
moon festival

cpfeiffer:
Thank you! I hope you had a good moon festival :-)

cpfeiffer:
Can you help us understand more of the authentication/initialization process?

AFAIU, there are multiple phases:
- first authentication where we send `0x1, 0x8` and the key to `00000009` characteristic
- then there is some kind of challenge response which we initiate by writing `0x2, 0x8` to `00000009` characteristic
- after getting the challenge, we encrypt it with the key and send `0x3, 0x8` and the encrypted challenge to `00000009` characteristic

But then we get `0x10,0x2,0x4` (when we did the first-time auth) or `0x10,0x3,0x4` (otherwise) and soon after that, the connection goes down. I guess that 0x4 means "unsuccessful".

If you can help us understand the process, I can help with with getting it implemented :-)

JohnnySun:
I think `0x10` means response, so if we receive `0x10,0x3,0x4` and connection goes down. maybe its unsuccessful. let me to some test to confirm.

JohnnySun:
ok , i found. `0x10, 0x3, 0x1' means auth succeful, and `0x10, 0x3, 0x4` means auth unsuccessful.

first, you need write `0x01, 0x08, 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40, 0x41, 0x42, 0x43, 0x44, 0x45` to `00000009` as key at band first connect to this app.

cpfeiffer:
It works! I have a stable connection. The problem was the sending of `0x2, 0x8` before receiving the `0x10,0x1,0x1`. Very cool. Needs some polish and cleanup, but at least the connection is there :-)

cpfeiffer:
I pushed my changes as ccdb843b6e2054875369dd88b347051bb1c2eac5 :-)

I moved the initialization code to class InitOperation for now, so that all that call-response state handling is separate. All that asynchronous writing queueing, event handling etc. is a bit ugly, I don't like it very much, but that's what we have at the moment, and it works.

I hope you do not have any local changes that are in conflict with my changes in this commit. If you have, feel free to ask me for help -- next time I should better coordinate this beforehand.

Thanks again!

PR #398, opened 2016-09-23
Add new BTLE API for Lollipop+ user

JohnnySun:
When discovery gadget use Lollipop+ device, use new BTLE API instead of old general BT discovery.


JohnnySun:
we can use List<ScanFilter> filters to filters our devices, at this time, i use `null` on filters

cpfeiffer:
Nice! I couldn't test your change yet, but it looks good to me. Could you change the ex.printStackTrace() to LOG instead? Just push the branch again, not need to open a new pull request.

Thanks! :thumbsup:

PR #443, opened 2016-12-01
Fix typos

uwehermann:
I've been browsing the source code for a bit, fixing random typos while doing so.

Completely untested, but it should not cause build or runtime issues (in theory).


cpfeiffer:
Very cool, thanks a lot!

cpfeiffer:
Wrong button :}

PR #446, opened 2016-12-05
Updated known issues for Mi Band

carlos-ferreira:
I was trying to "bond" the band with Gadgetbridge , but it only worked after unpairing from MiFit app and later trying again on Gadgetbridge.


cpfeiffer:
I haven't had such an issue myself. Which kind of Mi Band do you have (maybe we should add this to the text)?

carlos-ferreira:
I have a MI Band 1S.
This one http://www.gearbest.com/smart-wristband/pp_263275.html

cpfeiffer:
Alright, I'll merge it and rephrase it a bit. Thanks!

PR #453, opened 2016-12-08
Revamp Notification types Pebble

beschoenen:
This PR moves the big switches that handle the notification types for Pebble to the existing `NotificationType` `enum` and an `AppNotificationType` class which is a singleton `HashMap`. 

I hope that this is will make it easier in the future to add support for new apps.

photo_2016-12-08_19-22-40

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/NotificationListener.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/NotificationListener.java
@@ -23,20 +23,15 @@ import android.service.notification.StatusBarNotification;
 import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.RemoteInput;
 import android.support.v4.content.LocalBroadcastManager;
-
+import nodomain.freeyourgadget.gadgetbridge.GBApplication;
+import nodomain.freeyourgadget.gadgetbridge.model.*;
+import nodomain.freeyourgadget.gadgetbridge.service.DeviceCommunicationService;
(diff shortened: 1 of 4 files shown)


ashimokawa:
Thanks, that does not look too bad.
At first sight:

- put("com.moez.QKSMS", NotificationType.GENERIC_SMS); is duplicated
- please use "optimizie imports" and "reformat" code in android studio (if you use android studio), if we will do it later then it will expand your xyz.* imports again, so better do it before the merge
- please squash your commits when done ( I can also use the "squash and merge" function if you prefer that, just tell me)

beschoenen:
It might be a good idea to add a `.editorconfig` file, to keep the code more consistent.
http://editorconfig.org

ashimokawa:
No please only reformat the files you touched and only if you use android studio with default settings. It is not a requirement but I just wanted to mention it. 

The changes done on all the imports are worse because they will be changed back again by android studio (which we all use). 

Since we want to release really soon , I might just merge this as is this evening and then reformat before release, if you havent changed anything by then.

For the next time, the recommended way is to run "reformat code" before committing on VCS changed lines only (it is an option in android studio)

beschoenen:
Well, I use IntelliJ Idea not Android Studio and those import changes happened automatically, I didn't do them (I didn't even run a reformat).

ashimokawa:
That's weird since Android Studio is basically some customized intellij.

OK then I will merge as is. Would you mind if I squash the commits into one (it is a one click action in github , you authorship will be preserved)

beschoenen:
Sure, that's no problem

ashimokawa:
Thank you! Done.

PR #460, opened 2016-12-15
Set notification to minimal priority

ya-isakov:
Pebble app is not showing in status bar and on lockscreen. This change is for GadgetBridge to have the same behavior.


danielegobbetti:
Thanks for the contribution, I believe this should be controlled by a toggle in the user settings though, because personally I use the lockscreen notification all the time to be sure that the connection is present.
Can you please add this?

danielegobbetti:
Cherry picked.

PR #470, opened 2016-12-21
Add preliminary support for HPlus devices: Zeblaze Zeband and others

jpbarraca:
This seems to be working and can provide basic support for the Zeband. I would like some feedback and testing.
Tested on a Zeblaze Zeband
Not all protocol already decoded, but we will try to find the rest.

Working
- text and call notifications
- setting most user data, date and time
- heart rate monitoring (some flaws in continuous mode)

On the todo list
- sleep monitoring
- calorie monitoring
- distance monitoring

Probably feasible
- Alarms

Not clear right now
- Firmware update
- Battery status

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
@@ -56,9 +56,6 @@ public final class HPlusConstants {
     public static final byte COMMAND_SET_PREF_FINDME = 0x0a;
     public static final byte COMMAND_SET_PREF_SAVE = 0x17;
     public static final byte COMMAND_SET_PREF_END = 0x4f;
-    public static final byte COMMAND_SET_INCOMMING_SOCIAL = 0x31;
-    public static final byte COMMAND_SET_INCOMMING_SMS = 0x40;
-    public static final byte COMMAND_SET_DISPLAY_TEXT = 0x43;
     public static final byte COMMAND_SET_DISPLAY_ALERT = 0x23;
     public static final byte COMMAND_SET_PREF_ALLDAYHR = 53;
 
@@ -68,9 +65,17 @@ public final class HPlusConstants {
     public static final byte COMMAND_SET_CONF_SAVE = 0x17;
     public static final byte COMMAND_SET_CONF_END = 0x4f;
 
+    public static final byte COMMAND_SET_PREFS = 0x50;
+    public static final byte COMMAND_SET_SIT_INTERVAL = 0x51;
+
+
 
     public static final byte DATA_STATS = 0x33;
     public static final byte DATA_SLEEP = 0x1A;
+
+    public static final byte COMMAND_ACTION_INCOMING_SOCIAL = 0x31;
+    public static final byte COMMAND_ACTION_INCOMMING_SMS = 0x40;
+    public static final byte COMMAND_ACTION_DISPLAY_TEXT = 0x43;
 
 
     public static final String PREF_HPLUS_USER_ALIAS = "hplus_user_alias";
@@ -79,4 +84,11 @@ public final class HPlusConstants {
     public static final String PREF_HPLUS_ALLDAYHR = "hplus_alldayhr";
     public static final String PREF_HPLUS_UNIT = "hplus_unit";
     public static final String PREF_HPLUS_TIMEMODE = "hplus_timemode";
+    public static final String PREF_HPLUS_WRIST = "hplus_wrist";
+    public static final String PREF_HPLUS_SWALERT = "hplus_sw_alert";
+    public static final String PREF_HPLUS_ALERT_TIME = "hplus_alert_time";
+    public static final String PREF_HPLUS_SIT_START_TIME = "hplus_sit_start_time";
+    public static final String PREF_HPLUS_SIT_END_TIME = "hplus_sit_end_time";
+    public static final String PREF_HPLUS_COUNTRY = "hplus_country";
+
 }
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
@@ -72,6 +72,7 @@ public final class HPlusConstants {
 
     public static final byte DATA_STATS = 0x33;
     public static final byte DATA_STEPS = 0x36;
+    
     public static final byte DATA_SLEEP = 0x1A;
 
     public static final byte COMMAND_ACTION_INCOMING_SOCIAL = 0x31;
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
@@ -71,6 +71,7 @@ public final class HPlusConstants {
 
 
     public static final byte DATA_STATS = 0x33;
+    public static final byte DATA_STEPS = 0x36;
     public static final byte DATA_SLEEP = 0x1A;
 
     public static final byte COMMAND_ACTION_INCOMING_SOCIAL = 0x31;
@@ -78,8 +79,6 @@ public final class HPlusConstants {
     public static final byte COMMAND_ACTION_DISPLAY_TEXT = 0x43;
 
 
-    public static final String PREF_HPLUS_USER_ALIAS = "hplus_user_alias";
-    public static final String PREF_HPLUS_FITNESS_GOAL = "hplus_fitness_goal";
     public static final String PREF_HPLUS_SCREENTIME = "hplus_screentime";
     public static final String PREF_HPLUS_ALLDAYHR = "hplus_alldayhr";
     public static final String PREF_HPLUS_UNIT = "hplus_unit";
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
@@ -59,7 +59,7 @@ public final class HPlusConstants {
     public static final byte COMMAND_SET_DISPLAY_ALERT = 0x23;
     public static final byte COMMAND_SET_PREF_ALLDAYHR = 53;
 
-    public static final byte COMMAND_SET_INCOMMING_CALL = 65;
+    public static final byte COMMAND_SET_INCOMING_CALL = 0x41;
     public static final byte[] COMMAND_FACTORY_RESET = new byte[] {-74, 90};
 
     public static final byte COMMAND_SET_CONF_SAVE = 0x17;
@@ -75,8 +75,13 @@ public final class HPlusConstants {
     public static final byte DATA_SLEEP = 0x1A;
 
     public static final byte COMMAND_ACTION_INCOMING_SOCIAL = 0x31;
-    public static final byte COMMAND_ACTION_INCOMMING_SMS = 0x40;
+    public static final byte COMMAND_ACTION_INCOMING_SMS = 0x40;
     public static final byte COMMAND_ACTION_DISPLAY_TEXT = 0x43;
+    public static final byte[] COMMAND_ACTION_INCOMING_CALL = new byte[] {6, -86};
+    public static final byte COMMAND_ACTION_DISPLAY_TEXT_CENTER = 0x23;
+    public static final byte COMMAND_ACTION_DISPLAY_TEXT_NAME = 0x3F;
+    public static final byte COMMAND_ACTION_DISPLAY_TEXT_NAME_CN = 0x3E; //Text in GB2312?
+
 
 
     public static final String PREF_HPLUS_SCREENTIME = "hplus_screentime";
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
@@ -1,0 +1,82 @@
+package nodomain.freeyourgadget.gadgetbridge.devices.hplus;
+
+import java.util.UUID;
+
+/**
+ * Message constants reverse-engineered by João Paulo Barraca, jpbarraca@gmail.com.
+ *
+ * @author João Paulo Barraca &lt;jpbarraca@gmail.com&gt;
+ */
+public final class HPlusConstants {
+
+    public static final UUID UUID_CHARACTERISTIC_CONTROL = UUID.fromString("14702856-620a-3973-7c78-9cfff0876abd");
+    public static final UUID UUID_CHARACTERISTIC_MEASURE = UUID.fromString("14702853-620a-3973-7c78-9cfff0876abd");
+    public static final UUID UUID_SERVICE_HP = UUID.fromString("14701820-620a-3973-7c78-9cfff0876abd");
+
+
+    public static final byte COUNTRY_CN = 1;
+    public static final byte COUNTRY_OTHER = 2;
+
+    public static final byte CLOCK_24H = 0;
+    public static final byte CLOCK_12H = 1;
+
+    public static final byte UNIT_METRIC = 0;
+    public static final byte UNIT_IMPERIAL = 1;
+
+    public static final byte SEX_MALE = 0;
+    public static final byte SEX_FEMALE = 1;
+
+    public static final byte HEARTRATE_MEASURE_ON = 11;
+    public static final byte HEARTRATE_MEASURE_OFF = 22;
+
+    public static final byte HEARTRATE_ALLDAY_ON = 10;
+    public static final byte HEARTRATE_ALLDAY_OFF = -1;
+
+    public static final byte[] COMMAND_SET_INIT1 = new byte[]{0x50,0x00,0x25,(byte) 0xb1,0x4a,0x00,0x00,0x27,0x10,0x05,0x02,0x00,(byte) 0xff,0x0a,(byte) 0xff,0x00,(byte) 0xff,(byte) 0xff,0x00,0x01};
+    public static final byte[] COMMAND_SET_INIT2 = new byte[]{0x51,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x07,(byte) 0xe0,0x0c,0x12,0x16,0x0a,0x10,0x00,0x00,0x00,0x00};
+
+    public static final byte[] COMMAND_SET_PREF_START = new byte[]{0x4f, 0x5a};
+    public static final byte[] COMMAND_SET_PREF_START1 = new byte[]{0x4d};
+
+    public static final byte COMMAND_SET_PREF_COUNTRY = 0x22;
+    public static final byte COMMAND_SET_PREF_TIMEMODE = 0x47;
+    public static final byte COMMAND_SET_PREF_UNIT = 0x48;
+    public static final byte COMMAND_SET_PREF_SEX = 0x2d;
+
+    public static final byte COMMAND_SET_PREF_DATE = 0x08;
+    public static final byte COMMAND_SET_PREF_TIME = 0x09;
+    public static final byte COMMAND_SET_PREF_WEEK = 0x2a;
+    public static final byte COMMAND_SET_PREF_SIT = 0x1e;
+    public static final byte COMMAND_SET_PREF_WEIGHT = 0x05;
+    public static final byte COMMAND_SET_PREF_HEIGHT = 0x04;
+    public static final byte COMMAND_SET_PREF_AGE = 0x2c;
+    public static final byte COMMAND_SET_PREF_GOAL = 0x26;
+    public static final byte COMMAND_SET_PREF_SCREENTIME = 0x0b;
+    public static final byte COMMAND_SET_PREF_BLOOD = 0x4e; //??
+    public static final byte COMMAND_SET_PREF_FINDME = 0x0a;
+    public static final byte COMMAND_SET_PREF_SAVE = 0x17;
+    public static final byte COMMAND_SET_PREF_END = 0x4f;
+    public static final byte COMMAND_SET_INCOMMING_SOCIAL = 0x31;
+    public static final byte COMMAND_SET_INCOMMING_SMS = 0x40;
+    public static final byte COMMAND_SET_DISPLAY_TEXT = 0x43;
+    public static final byte COMMAND_SET_DISPLAY_ALERT = 0x23;
+    public static final byte COMMAND_SET_PREF_ALLDAYHR = 53;
+
+    public static final byte COMMAND_SET_INCOMMING_CALL = 65;
+    public static final byte[] COMMAND_FACTORY_RESET = new byte[] {-74, 90};
+
+    public static final byte COMMAND_SET_CONF_SAVE = 0x17;
+    public static final byte COMMAND_SET_CONF_END = 0x4f;
+
+
+    public static final byte DATA_STATS = 0x33;
+    public static final byte DATA_SLEEP = 0x1A;
+
+
+    public static final String PREF_HPLUS_USER_ALIAS = "hplus_user_alias";
+    public static final String PREF_HPLUS_FITNESS_GOAL = "hplus_fitness_goal";
+    public static final String PREF_HPLUS_SCREENTIME = "hplus_screentime";
+    public static final String PREF_HPLUS_ALLDAYHR = "hplus_alldayhr";
+    public static final String PREF_HPLUS_UNIT = "hplus_unit";
+    public static final String PREF_HPLUS_TIMEMODE = "hplus_timemode";
+}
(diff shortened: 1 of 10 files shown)


cpfeiffer:
Wow, excellent job! Will have a detailed look ASAP.

ashimokawa:
@jpbarraca 
Very cool! Thanks.
Unfortunately no one of us owns that device. I still would like to merge that soon. More devices are always welcome and that is our first external new device contribution!

Oh and one question: You write "Zeblaze Zeband and others". Which would be the others? Have you tested any?

jpbarraca:
Do not hurry. But please provide feedback. I'm struggling to find how to actually store data. Hearth Rate seems to be easy. But other things are not obvious to me.
These watches support Calories, Distance, Steps, HR, Paddles (swimming) and Sleep tracking.

I'm not sure how many HPlus devices there are. In the http://hpluswatch.com site there are several, then you have the zeband. I suspect that the Smart F68 (Makibes?) watch may be compatible.

cpfeiffer:
Will do so in a few hours.

cpfeiffer:
First off, you've done one helluva good job!

Is the additional data provided per sample or already aggregated? If it's per sample, we need to extend ActivitySample and the implementations. We do not have built-in support for pre-aggregated data yet. At the moment, we calculate everything from the individual samples.

We are very flexible regarding the database tables, though. Every device can have its own tables and thus store its own data. We should try to get a common interface though, that other devices with similar capabilities can share.

So, everything that is related to samples belongs to ActivitySample and its implementations. GBDaoGenerator creates the device specific entities that (among others) extend AbstractActivitySample.

SampleProvider is the common interface for reading and writing the samples, with device specific implementations.

If you have further questions, please ask :-) I'll now review the code and add comments in-place.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
```diff
@@ -0,0 +1,236 @@
...
+
+public class HPlusCoordinator extends AbstractDeviceCoordinator {
+    private static final Logger LOG = LoggerFactory.getLogger(HPlusCoordinator.class);
+    private static Prefs prefs  = GBApplication.getPrefs();
+
+    @Override
+    public DeviceType getSupportedType(GBDeviceCandidate candidate) {
+        String name = candidate.getDevice().getName();
+        LOG.debug("Looking for: " + name);
+        if (name != null && name.startsWith("HPLUS")) {
```
cpfeiffer: Is this sufficient for you? I know that checking the device is far from perfect but that's all we have right now.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
```diff
@@ -0,0 +1,236 @@
...
+        return prefs.getInt(HPlusConstants.PREF_HPLUS_FITNESS_GOAL + "_" + address, 10000);
+    }
+
+    /**
+     * Returns the user info from the user configured data in the preferences.
+     *
+     * @param hplusAddress
+     * @throws IllegalArgumentException when the user info can not be created
+     */
+    public static UserInfo getConfiguredUserInfo(String hplusAddress) throws IllegalArgumentException {
```
cpfeiffer: That's not needed, is it? (A Mi Band leftover, I suppose)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusSampleProvider.java
```diff
@@ -0,0 +1,140 @@
...
+import nodomain.freeyourgadget.gadgetbridge.entities.DaoSession;
+import nodomain.freeyourgadget.gadgetbridge.entities.Device;
+import nodomain.freeyourgadget.gadgetbridge.entities.HPlusHealthActivitySample;
+import nodomain.freeyourgadget.gadgetbridge.entities.HPlusHealthActivitySampleDao;
+import nodomain.freeyourgadget.gadgetbridge.impl.GBDevice;
+import nodomain.freeyourgadget.gadgetbridge.model.ActivityKind;
+
+public class HPlusSampleProvider extends AbstractSampleProvider<HPlusHealthActivitySample> {
+
+    public static final int TYPE_DEEP_SLEEP = 4;
```
cpfeiffer: These values are the same as those in MiBandSampleProvider. Are you sure HPLUS uses the same values?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusSampleProvider.java
```diff
@@ -0,0 +1,140 @@
...
+                return TYPE_NONWEAR;
+            case ActivityKind.TYPE_UNKNOWN: // fall through
+            default:
+                return TYPE_UNKNOWN;
+        }
+    }
+
+
+    @Override
+    public List<HPlusHealthActivitySample> getAllActivitySamples(int timestamp_from, int timestamp_to) {
```
cpfeiffer: I think here's something that can be improved. Could it be that you didn't get any samples with super.getAllActivitySamples() because of the (possibly wrong) activity-kind constants/mapping above?

If the constants and the mapping with normalizeType() and toRawActivityKind() works, you shouldn't need to reimplement the query.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSleepRecord.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Calendar;
+
+
+
+public class HPlusSleepRecord {
+    private long bedTimeStart;
```
cpfeiffer: Could you add a comment about the time units? Every device has its own kind of units (like e.g. year - 2000 for the Mi Band 1).

cpfeiffer:
Hm, I think github showed me only parts of your PR. Will review the rest in a bit.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
```diff
@@ -0,0 +1,236 @@
...
+    public static byte getUserWeight(String address) {
+        ActivityUser activityUser = new ActivityUser();
+
+        return (byte) activityUser.getWeightKg();
+    }
+
+    public static byte getUserHeight(String address) {
+        ActivityUser activityUser = new ActivityUser();
+
+        return (byte) activityUser.getHeightCm();
```
cpfeiffer: Here you might need to ensure unsigned value with & 0xff. You basically need to do this everywhere a value > 128 is used.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+
+                return new byte[]{
+                        HPlusConstants.COMMAND_SET_PREF_COUNTRY,
+                        sex,
+                        age,
+                        bodyHeight,
+                        bodyWeight,
+                        0,
+                        0,
+                        (byte) (goal / 256),
```
cpfeiffer: & 0xff? (Imagine someone training for marathon)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+            @Override
+            protected byte[] checkCondition() {
+                int startTime = HPlusCoordinator.getSITStartTime(getDevice().getAddress());
+                int endTime = HPlusCoordinator.getSITEndTime(getDevice().getAddress());
+
+                Calendar now = Calendar.getInstance();
+
+                return new byte[]{
+                        HPlusConstants.COMMAND_SET_SIT_INTERVAL,
+                        (byte) (startTime / 256),
```
cpfeiffer: & 0xff?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+    }
+
+
+    private HPlusSupport setFindMe(TransactionBuilder transaction) {
+        LOG.info("Attempting to set Findme...");
+        return this;
+    }
+
+    private HPlusSupport requestDeviceInfo(TransactionBuilder builder) {
+        LOG.debug("Requesting Device Info!");
```
cpfeiffer: Does it perhaps implement the Bluetooth Device Info Profile? See class DeviceInfoProfile.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+    @Override
+    public void onNotification(NotificationSpec notificationSpec) {
+        LOG.debug("Got Notification");
+        showText(notificationSpec.body);
+    }
+
+
+    @Override
+    public void onSetTime() {
+        TransactionBuilder builder = new TransactionBuilder("vibration");
```
cpfeiffer: "vibration"? :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+    }
+
+    @Override
+    public void onFindDevice(boolean start) {
+        LOG.debug("Find Me");
+
+        getQueue().clear();
+        ctrlCharacteristic = getCharacteristic(HPlusConstants.UUID_CHARACTERISTIC_CONTROL);
+
+        TransactionBuilder builder = new TransactionBuilder("findMe");
```
cpfeiffer: Typically we use performInitialized() to acquire a TransactionBuilder(). This will attempt to connect and initialize the device prior to executing the action (here: "find device").

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+
+        try (DBHandler handler = GBApplication.acquireDB()) {
+            DaoSession session = handler.getDaoSession();
+
+            Device device = DBHelper.getDevice(getDevice(), session);
+            User user = DBHelper.getUser(session);
+            int ts = (int) (System.currentTimeMillis() / 1000);
+            HPlusSampleProvider provider = new HPlusSampleProvider(gbDevice, session);
+
+            //TODO: Store Sample. How?
```
cpfeiffer: You can store sleep information the same way as in the method below. I'd buffer the samples locally before committing them to the database. This would allow combining activity samples with sleep, but this depends on how the device sends the data.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    public static byte getUnit(String address) {
+        return (byte) prefs.getInt(HPlusConstants.PREF_HPLUS_UNIT + "_" + address, 0);
+    }
+
+    public static byte getUserWeight(String address) {
+        ActivityUser activityUser = new ActivityUser();
+
+        return (byte) activityUser.getWeightKg();
```
cpfeiffer: & 0xff?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+    private HPlusSupport setTimeMode(TransactionBuilder transaction) {
+        LOG.info("Attempting to set Time Mode...");
+        transaction.add(new ConditionalWriteAction(ctrlCharacteristic) {
+            @Override
+            protected byte[] checkCondition() {
+
+                byte value = HPlusCoordinator.getTimeMode(getDevice().getAddress());
+                return new byte[]{
+                        HPlusConstants.COMMAND_SET_PREF_TIMEMODE,
+                        (byte) value
```
cpfeiffer: There's quite a few superfluous casts, probably because the "value" variable was of type int before.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
jpbarraca: I have two Zebands and they do not share the same MAC prefix. I have tried filtering by the existence of a particular service and it worked ok, so I will be add this additional check.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
jpbarraca: Yes. Removed.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusSampleProvider.java
jpbarraca: These bands have not predefined raw types and according to what I found, they only allow determining if the user is awake or sleeping. Will remove this.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+
+        //Initialize device
+        setInitValues(builder);
+        syncPreferences(builder);
+
+        builder.notify(getCharacteristic(HPlusConstants.UUID_CHARACTERISTIC_MEASURE), true);
+
+        UUID uuid = UUID.fromString("00002902-0000-1000-8000-00805f9b34fb");
+        BluetoothGattDescriptor descriptor = measureCharacteristic.getDescriptor(uuid);
+        descriptor.setValue(BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE);
```
jpbarraca: I was not sure if the building would do it. It was probably a leftover.
I removed this, tested and notifications still work.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -0,0 +1,881 @@
...
+        syncPreferences(builder);
+
+        builder.write(ctrlCharacteristic, new byte[]{HPlusConstants.COMMAND_SET_CONF_SAVE});
+        builder.write(ctrlCharacteristic, new byte[]{HPlusConstants.COMMAND_SET_CONF_END});
+        return this;
+    }
+
+    private HPlusSupport syncPreferences(TransactionBuilder transaction) {
+        LOG.info("Attempting to sync preferences...");
+        transaction.add(new ConditionalWriteAction(ctrlCharacteristic) {
```
jpbarraca: Refactored all code and tested. Will commit changes.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: Marathon runners will have to use better sports bands ;) Added the & 0xff.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: Cleaning these

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: fixed all occurrences.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: No device info or battery info :/
Supports: Generic Access (0x1800), Generic Attribute (0x1801), HpService and DFU Service

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: Corrected

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: OK

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
jpbarraca: Fixed the pattern in all places.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusCoordinator.java
jpbarraca: Fixed the pattern in all places

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: That arrives fragmented and with different periodicities.
The device also provides instant data and aggregates per 10m, and per day. When I have a better notion of all messages supported and their format I will check this approach.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusSampleProvider.java
jpbarraca: Ok. Will remove this for now.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/hplus/HPlusConstants.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public static final byte DATA_STATS = 0x33;
+    public static final byte DATA_SLEEP = 0x1A;
+
+    public static final byte COMMAND_ACTION_INCOMING_SOCIAL = 0x31;
+    public static final byte COMMAND_ACTION_INCOMMING_SMS = 0x40;
+    public static final byte COMMAND_ACTION_DISPLAY_TEXT = 0x43;
+
+
+    public static final String PREF_HPLUS_USER_ALIAS = "hplus_user_alias";
```
jpbarraca: I am all in favor of sharing these preferences, as some are useful to multiple devices, and if you find appropriate I can add it that way.

jpbarraca:
Currently the code has fully working text notifications to the device. I have been testing it in standard use and it seems to be ok.

I would suggest going with this, and later integrate the code that handles and stores the data reported by the device.

Most of the code is copied/pasted (and then refactored) from GadgetBridge itself and should follow the same license.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -653,11 +653,11 @@ private void showText(String title, String body) {
             String message = "";
 
             if (title != null) {
-                if (title.length() > 12) {
+                if (title.length() > 17) {
                     message = title.substring(0, 12);
```
cpfeiffer: Really 12?

jpbarraca: :) fixed

cpfeiffer: AFAIU you're trying to create a message that is exactly n characters long (so that it fits onto the display. Maybe we should have a method (or two) in a class StringUtils (or similar) that does this:
truncate(String, numCharacters)
pad(String, numCharacters)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -233,7 +236,7 @@ private HPlusSupport setCurrentDate(TransactionBuilder transaction) {
                 HPlusConstants.COMMAND_SET_PREF_DATE,
                 (byte) ((year / 256) & 0xff),
                 (byte) (year % 256),
-                (byte) (month),
+                (byte) (month + 1),
```
cpfeiffer: I actually thought that you might need the offset, but I thought you'd know if the device shows the correct date :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -636,6 +636,79 @@ public void onTestNewFunction() {
...
+            builder.write(ctrlCharacteristic, new byte[] {HPlusConstants.COMMAND_SET_INCOMING_CALL, 1 });
+
+            //Show Call Icon
+            builder.write(ctrlCharacteristic, HPlusConstants.COMMAND_ACTION_INCOMING_CALL);
+
+            //builder = performInitialized("incomingCallText");
+            builder.queue(getQueue());
+
+            try {
+                Thread.sleep(200);
```
cpfeiffer: We try hard to avoid such timing issues. Why do you need to sleep here? Do you have to wait for a response of the device before you can send the next command?

So far we usually do this with "Operations" that are registered as listeners and can react to gatt characteristic change events. For simple "confirmation" responses I'd like to simplify this so that we can avoid creating an operation, but I haven't done this yet.

However, instead of Thread.sleep(), you can add a `WaitAction` to the queue (and thus use the same transaction for the whole stuff).

jpbarraca: I will use the WaitAction. Apparently the device will loose messages if they are sent without delay. There is no feedback to trigger a new message.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: The month number is small in the display and I missed the error :)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
jpbarraca: Makes sense as it can be reused. Will add the TODO

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
cpfeiffer: This shouldn't be a problem. You have total freedom over how you store sample data. Your custom SampleProvider can then figure out the details to provide the samples in a "normalized" way so that the charts can display them.

E.g. some Pebble Health data is stored as some kind of "overlay" data and combined with the other data by the sample provider.

However, if you realize that this does not make sense for HPlus, we can also provide some kind of aggregate data interface. Needs some thinking because we don't have something like that yet.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
cpfeiffer: Have a look at the two `FetchActivityOperation` classes. For Mi1 you just ask to sync and the band tells you how much data you will get. For Mi2, you explicitly send a data from which you want to receive activity data. We look up the last sample in the database and use that timestamp. If there is none, we simply use current date - 10 days, IIRC.

Did you try sending the commands without the sleep()? I've seen other implementations using sleep() because they don't have proper synchronization of the BLE commands.

cpfeiffer:
Send a note when you want us to merge it. From my POV it is ready for that.

jpbarraca:
@cpfeiffer Merge it. I will create a new branch to work on the data handling and storage.

cpfeiffer:
There we go! Great job and happy Christmas :-)

PR #489, opened 2017-01-03
Show K9 message body in notifications

hackoder:
K9 message body was not showing up in notifications since 0.15.0. This PR fixes that.


danielegobbetti:
Thanks!
Do you perhaps know if it would be wise to differentiate between old versions of K9 (<5.200) or if we can do this safely for every version?
I haven't checked their code wrt how the notifications are built.

hackoder:
I took a quick look at k9's notification code: https://github.com/k9mail/k-9/blob/master/k9mail/src/main/java/com/fsck/k9/notification/DeviceNotifications.java

Doesn't seem like it has changed much in past versions.

Issue #301, opened 2016-05-12
More Intuitive User Interface


vext01:
Hey,

I would like to start a discussion about possible changes to the UI of Gadgetbridge.

BTW, I know very little as to what you actually get off the device, so apologies if some of this can't be helped.

Let's jump in.

(Does anyone know how to scale down images on github?)

# Your Activity and Sleep

![quickmemo _2016-05-12-21-21-12](https://cloud.githubusercontent.com/assets/604955/15229642/ddc7b7be-1889-11e6-9db0-7080b07af009.png)

The page shows activity and sleep over time.

 * What is unit of the the Y-axis? I think the ticks are corresponding to a unit of heart rate (if you zoom the graph in, there are little numbers that match the Y-axis). So is that BPM? Axes labels would help.
 * What about the green bits? What scale are they on? Are they raw step samples collected in a fixed time period? If so, that's probably not a useful metric for users. How about a cumulative step count over time for the current day, a bit like you see on the android battery usage screen?
 * If the green bars and the yellow line are on different scales, how about adding more y-axis labels on the left with green numbers, the existing y-axis on the right could be coloured yellow.
 * Why is there only heart-rate data for sleep?

Another visualisation that might be cool is the one from the LG Health app. It looks like this:

![quickmemo _2016-05-12-21-22-37](https://cloud.githubusercontent.com/assets/604955/15230202/6b93f7f4-188c-11e6-92c0-bc964b791f45.png)

# Your Sleep

![quickmemo _2016-05-12-21-21-36](https://cloud.githubusercontent.com/assets/604955/15229943/31feacd8-188b-11e6-84c6-46b453a1019c.png)

 * Similarly, add axes labels.
 * Text in the pie chart is very small.

# Steps a week

![quickmemo _2016-05-12-21-51-23](https://cloud.githubusercontent.com/assets/604955/15230062/d7c502e8-188b-11e6-8189-3cbf4fe61035.png)

This page is fine, but I must admit I like the "LG health" interface more. It looks like this:

![quickmemo _2016-05-12-21-22-55](https://cloud.githubusercontent.com/assets/604955/15230111/1395598a-188c-11e6-8302-1add4480b098.png)

You can click on the days to zoom in and get more info. You have heart-rate data, so you could do something similar for that too if you wanted.

# Live Activity

 * "start your activity" looks like a pressable button. It isn't ;)

# Sidebar?

[...]

PR #491, opened 2017-01-04
HPlus: Improves device support

jpbarraca:
This will improve the support for HPlus devices by allowing data synchronization (HR, steps and sleep) along the following features:
* Discovery and pairing (improved)
* Notifications (Display + vibration) (improved)
* Synchronize the time and date (added)
* Display firmware version (added)
* Display battery charge (added)
* Near Real time Heart rate measurement (added)
* Synchronize activity data (alpha): only sleep, steps and intensity.
* Improved disconnection issues

Charts will be fully functional except for activity lines (working on it) and real time steps (no direct support in the band for this but, can probably be extrapolated from distance counters)

Calories, distance and individual sleep phases stored but not presented.

User preferences used whenever possible but the configuration of some parameters is still missing (Language, HR thresholds, Inactivity alarms, Real Time HR, SMS and Phone notifications, Find Me, Goals, Screen Time, ...)

The database format for the device was modified in order to include overlays.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordSteps.java
```diff
@@ -0,0 +1,48 @@
...
+/*
+* @author João Paulo Barraca &lt;jpbarraca@gmail.com&gt;
+*/
+
+
+import java.util.Calendar;
+
+
+class HPlusDataRecordSteps extends HPlusDataRecord{
+    public int steps;
```
cpfeiffer: (Can never forget https://en.wikipedia.org/wiki/Mars_Climate_Orbiter#Cause_of_failure)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
```diff
@@ -0,0 +1,446 @@
...
+import nodomain.freeyourgadget.gadgetbridge.entities.HPlusHealthActivityOverlayDao;
+import nodomain.freeyourgadget.gadgetbridge.entities.HPlusHealthActivitySample;
+import nodomain.freeyourgadget.gadgetbridge.impl.GBDevice;
+import nodomain.freeyourgadget.gadgetbridge.model.ActivityKind;
+import nodomain.freeyourgadget.gadgetbridge.model.ActivitySample;
+import nodomain.freeyourgadget.gadgetbridge.service.btle.TransactionBuilder;
+import nodomain.freeyourgadget.gadgetbridge.service.serial.GBDeviceIoThread;
+
+
+class HPlusHandlerThread extends GBDeviceIoThread {
```
cpfeiffer: The IoThread class is usually only used for the Bluetooth Classic (serial) protocol. Why do you need to use a separate thread? The use of BtLEQueue already makes it everything async.

Is there a reason why you can't implement something like a FetchOperation?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
```diff
@@ -0,0 +1,446 @@
...
+        if(record.heartRate == 255) {
+            getDevice().setFirmwareVersion2("---");
+            getDevice().sendDeviceUpdateIntent(getContext());
+            return true;
+        }
+
+        getDevice().setFirmwareVersion2("" + record.heartRate);
+        getDevice().sendDeviceUpdateIntent(getContext());
+
+        try (DBHandler dbHandler = GBApplication.acquireDB()) {
```
cpfeiffer: Is it possible to buffer the samples locally and only commit them to the db when finished (or some buffer threhold is reached)? Inserting every sample separate is a little costly, performance wise.

jpbarraca: Please correct me :)

The device will report Heart Rate periodically. Each message is independent of the previous messages, so I do not see that we reach an end. Also, this data is not stored in the device. Either we store it or it can be lost. I already filter data so that we only store unique packets.

Day and Sleep summaries are stored in the device. However, they are only updated once per day (each).

Day details are sent in slots, with 144 slots per day. This can be buffered and then stored.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordSteps.java
```diff
@@ -0,0 +1,48 @@
...
+            /*
+            unknown fields
+            short s12 = (short)(data[5] + data[6] * 256);
+            short s13 = (short)(data[7] + data[8] * 256);
+            short s16 = (short)(data[13]) + data[14] * 256);
+            short s17 = data[15];
+            short s18 = data[16];
+            */
+
+        Calendar date = Calendar.getInstance();
```
jpbarraca: Forgot that detail. 
PR is updated.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordSteps.java
jpbarraca: PR updated with comments to all message fields

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
jpbarraca: Actually this is far from perfect as events will be randomly skewed, and exceptions can break everything. Similar to what happens with a Timer.

This single thread will take care of some operations that should be done periodically, and with variable periodicity.
- Send a periodic keep alive packet ~60s
- Fetch Current Day data. This a burst of synchronous messages (1 get, 6 answers) plus a request/notification every 10 minutes
- Fetch Day Summaries ~ once per day
- Fetch Sleep Statistics ~ twice per day

I looked for an alternative way but this was the simplest I found, and saw no support for scheduling of commands in GB. Can you provide some pointers?

jpbarraca:
Found a bug with the total steps per day (overcounting).
Please do not merge.

cpfeiffer:
Alright.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
cpfeiffer: OK, you know the details, not me, so maybe my suggestion was rubbish :-)

I thought you would get a bunch a samples in one batch, so you could buffer them during the batch and, once that batch is transferred, commit it to the database.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
cpfeiffer: You're totally right, we're lacking such a thing. I'd love to have some generic support for such things though. We definitely need it for automatic activity sync with other devices as well.

I had imagined to implement something a bit like anacron though. Interested parties configure     a time/interval and the anacron service invokes all clients accordingly. It would use `AlarmManager` to wake up when necessary, and it would attempt to batch invocations in order to minimize wakeups.

It would also handle failures (e.g. when a device is not connected) and attempt to repeat the invocation.

Haven't gotten around to implementing that, though.

cpfeiffer: Is the keep-alive really necessary? I don't think we have such a thing for the Pebble or for the Liveview.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
jpbarraca: That happens when initially synchronizing, but then we receive each message independently.
The suggestion was valid and I will try to apply the advice as much as it seems reasonable.

jpbarraca:
All data being reported and displayed correctly. Screenshots attached.

What is not working (or properly tested):
* Inactivity alarms
* Alarms
* Swimming statistics
* Configuration of band specific settings. Some of these are not that specific and could be in the UserSettings.

Messages not parsed
* Messages of type 77: unknown purpose.
* 5m interval summaries: unknown purpose as data is inconsistent. 10m historical intervals (same message format) are processed.

hplus1
hplus2
hplus3
hplus4
hplus5

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
jpbarraca: The capability to keep timers (but not Java Timers) would be great.
Also, the SampleProviders could have more specific methods like: getTodaySteps, getWeekStepSummary, etc... That could improve performance of the provider. In order to keep previous behaviour.

Now that the code is more stable I will try disabling the keep alive packets and see what happens.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
```diff
@@ -330,66 +342,65 @@ public boolean processIncomingSleepData(byte[] data){
...
-        if(record.heartRate == 255) {
+        //Skip when measuring heart rate
+        //Calories and Distance are updated and these values will be lost.
+        //Because a message with a valid Heart Rate will be provided, this loss very limited
+        if(record.heartRate == ActivityKind.TYPE_NOT_MEASURED) {
             getDevice().setFirmwareVersion2("---");
             getDevice().sendDeviceUpdateIntent(getContext());
-            return true;
+        }else {
+            getDevice().setFirmwareVersion2("" + record.heartRate);
```
cpfeiffer: Nice trick to display the heart rate :smile:

We definitel need a UI to show current heart rate somewhere.

cpfeiffer: BTW, there's GBDevice.addDeviceInfo() that you can use to show additional information in the same way.

jpbarraca: Actually this is ugly... as I am putting the Heart Rate there.
If you prefer I may use that method. Or remove this altogether.

I could be interesting to have current device stats like battery available in the main interface. For this device I would be interested in showing Total Steps, Distance, Heart Rate, Calories and Battery.

jpbarraca: There could be an area with a table layout where each cell contains two areas (Title and Value) and devices could load a list of items to put there. This would be dynamic and flexible enough to show many things.

cpfeiffer: That's pretty much how it should be already if you use addDeviceInfo(). You can add all your desired fields like that. Do you want to show this information directly, instead of collapsed by default?

jpbarraca: IMHO this was an hack as I think GB will probably improve the way real time data is presented.

cpfeiffer: Yes, for sure we need a better way :-) If you have some ideas, you can add them to #301. But for now, you can display all your data with addDeviceInfo().

Ziw18:
Hey. Bracelet HPlus.
With an incoming call, the bracelet vibrates 3 times and the phone icon is displayed on the screen. After that, a short pause. And again vibrates 3 times with the caller's name displayed on the screen.

This is very inconvenient - you need to wait to find out who is calling.

Please take a look: when the call was received, both the handset icon and the caller's name were displayed at the same time. Let the pictogram be smaller and the name will be shorter, but it will be much more convenient.

jpbarraca:
We do not control the size of the pictogram and these delays are imposed by the bracelet. GB triggers two events: incoming call and set callee number. The bracelet does the rest.

A possibility would be to reverse the events (if the bracelet allows it)

Ziw18:
How much I correctly understood - it is possible to make, that at an incoming call the name of the caller was displayed only? Without any pictograms.If so, it would be very convenient.   07.03.2018, 02:55, "João Paulo Barraca" <notifications@github.com>:We do not control the size of the pictogram and these delays are imposed by the bracelet. GB triggers two events: incoming call and set callee number. The bracelet does the rest.A possibility would be to reverse the events (if the bracelet allows it)—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or mute the thread.

jpbarraca:
Please open an issue so that this is discussed as an Usability aspect

PR #495, opened 2017-01-06
Android automatic refactor - ObsoleteLayoutParam

luiscruz:
Hi,

I am developing a tool to automatically refactor Android applications with the goal of improving energy efficiency.
This pull request has the changes generated while applying the rule "ObsoleteLayoutParam".

While developing your application's views you might be specifying attributes in a view's artefact that are not necessary due to the nature of its parent. In this PR, those attributes were replaced by a comment.

I have made a previous validation of the changes and they seem correct.
Unfortunately, this tool is not able keep the original whitespace of the files, so comparison without ignoring whitespace might be confusing.
Please consider the changes and let me know if you agree with them.

Best,
Luis

### DIFF
--- a/app/src/main/res/layout/activity_configure_alarms.xml
+++ b/app/src/main/res/layout/activity_configure_alarms.xml
@@ -1,16 +1,19 @@
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools" android:layout_width="match_parent"
-    android:layout_height="match_parent" android:paddingLeft="@dimen/activity_horizontal_margin"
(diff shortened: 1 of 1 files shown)


cpfeiffer:
Interesting tool -- do you have a list of supported refactorings? I couldn't find one during a quick glance. And thanks for the pull request :-)

luiscruz:
Hi @cpfeiffer I realised now that I haven't had the change to reply in this thread.
You can read mor about the tool and supported refactorings here: https://luiscruz.github.io/papers/cruz2018using.pdf

Also, you may find this catalog interesting: https://tqrg.github.io/energy-patterns/

Cheers!

PR #500, opened 2017-01-13
Transliterate unsupported Russian characters into English

ivanovlev:
Zeblaze ZeBand don't support russian language, so i found the great solution. Just transliterated to eng unsupported chars. Other symbols are not affected.

Method can be extended for other languages, it's very easy.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,6 +77,7 @@ dependencies {
     compile 'org.apache.commons:commons-lang3:3.4'
 
 //    compile project(":DaoCore")
+    compile 'com.google.guava:guava:19.0'
 }
 
 preBuild.dependsOn(":GBDaoGenerator:genSources")
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,7 +77,6 @@ dependencies {
     compile 'org.apache.commons:commons-lang3:3.4'
 
 //    compile project(":DaoCore")
-    compile 'com.google.guava:guava:19.0'
 }
 
 preBuild.dependsOn(":GBDaoGenerator:genSources")
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceCommunicationService.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceCommunicationService.java
@@ -51,6 +51,7 @@ import nodomain.freeyourgadget.gadgetbridge.model.WeatherSpec;
 import nodomain.freeyourgadget.gadgetbridge.util.DeviceHelper;
 import nodomain.freeyourgadget.gadgetbridge.util.GB;
 import nodomain.freeyourgadget.gadgetbridge.util.GBPrefs;
+import nodomain.freeyourgadget.gadgetbridge.util.LanguageUtils;
 import nodomain.freeyourgadget.gadgetbridge.util.Prefs;
 
 import static nodomain.freeyourgadget.gadgetbridge.model.DeviceService.ACTION_ADD_CALENDAREVENT;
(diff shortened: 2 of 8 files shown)


cpfeiffer:
Cool, thanks for the fix and congratulations for getting Issue #500 :-)

Would you mind moving the transliteration functionality into a separate class, e.g. LanguageUtils in the utils package? Then we can easily reuse it for other devices with similar problems.

And we can spare the additional guava dependency by using Collections.unmodifiableMap() instead.

jpbarraca:
Very nice. I have the same problem with Portuguese.

ashimokawa:
Are ä,ö,ü and ß also unsupported?

jpbarraca:
Actually, the band displays the ASCII table correctly with some messages, but with the current code, accented letters do not work. I have problems with á õ ã à é í ó ç â ê and so on.

ivanovlev:
cpfeiffer, ok

ivanovlev:
jpbarraca, afformative, after 175 just china letters. Bitches

but á õ ã à é í ç â ê still present, except ó. I'm working under specific overrides  by ascii codes for standard transliterations...

ivanovlev:
Now orčpžsíáýd will be conveted into orcpzsiayd

jpbarraca:
The band expects messages to be in GB2312.

ashimokawa:
@jpbarraca 
But no complete charset I guess, so still no cyrillic?

http://www.herongyang.com/gb2312/GB2312-to-Unicode-Row-07-Cyrillic-Character.html

jpbarraca:
It seems that the charset is limited to 8 bits.

cpfeiffer:
Did you have a look at http://www.herongyang.com/gb2312/GB2312Unicode-Java-GB2312-Unicode-Mapping.html ?
Among others, it says
> // GB2312 is not supported by JDK. So I am using GBK.

So with the encoding class given above, we could properly encode the strings instead of just casting char to byte. (See String.getBytes(Charset))

jpbarraca:
Had no luck with characters above 175. Even when sending specific byte sequences using a dedicated python script. The original app also has problems with accented characters.

This lets me believe that the charset is limited.

ashimokawa:
@jpbarraca 

Couldn't you reproduce what @ivanovlev said? I mean Chinese characters above 175.
If there are Chinese characters, it seems unlikely to me that it is an 8-bit encoding. If it is still an 8-bit encoding, the chinese characters could be only the ones used for time and date like 月火水木金土日年. Unfortunately, lacking the device, I cannot test..

ivanovlev:
If i modify debug screen command "Send" for codes "String.valueOf(Character.toChars(int))", the device can correctly show chars under to 175, 176 return black, 177+ returns china letters. ~250 again returns black, after 255 table repeats (a,b,c...)

cpfeiffer:
So what's the current status of this PR -- should we merge it or are you still working on the details of encoding?

ivanovlev:
Yes you can merge, i thought about extended russian transliteration (set different replacement current character by previous character), but I will not do it. Method becomes a to closed relative the other languages.

cpfeiffer:
Thanks :-)

ivanovlev:
@cpfeiffer Can you change one missing thing - Debug > "Send notification". I just seen, that debug "Send" button doesn't use DeviceCommunicationService, so the transliteration does not apply.

`String testString = editContent.getText().toString();`

to

`String testString = editContent.getText().toString();

if(LanguageUtils.transliterate())
{
        testString = LanguageUtils.transliterate(testString);
}
`

ivanovlev:
@cpfeiffer Can you change one missing thing - Debug > "Send notification". I just seen, that debug "Send" button doesn't use DeviceCommunicationService, so the transliteration does not apply.

DebugActivity.java

add simple method

` // Return user input from screen`
`    private String GetUserMessage()`
`    {`
`        String message = editContent.getText().toString();`
`        if (LanguageUtils.transliterate())`
`        {`
`            return LanguageUtils.transliterate(message);`
`        }`

`        return message;`
`    }`

and replace all usage of

`String testString = editContent.getText().toString();`

to `GetUserMessage()`

PR #513, opened 2017-01-20
Simplification of transliteration integration

ivanovlev:
In process of discussion, first target was chaged to simple refactoring

> I just seen, that debug "Send" button doesn't use DeviceCommunicationService, so the transliteration does not apply. This is fix for it, manual call..

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
@@ -180,9 +180,10 @@ public class DebugActivity extends GBActivity {
             @Override
             public void onClick(View v) {
                 MusicSpec musicSpec = new MusicSpec();
-                musicSpec.artist = editContent.getText().toString() + "(artist)";
-                musicSpec.album = editContent.getText().toString() + "(album)";
-                musicSpec.track = editContent.getText().toString() + "(track)";
+                String testString = editContent.getText().toString();
+                musicSpec.artist = testString + "(artist)";
+                musicSpec.album = testString + "(album)";
+                musicSpec.track = testString + "(track)";
                 musicSpec.duration = 10;
                 musicSpec.trackCount = 5;
                 musicSpec.trackNr = 2;
--- a/app/src/test/java/nodomain/freeyourgadget/gadgetbridge/service/TestDeviceService.java
+++ b/app/src/test/java/nodomain/freeyourgadget/gadgetbridge/service/TestDeviceService.java
@@ -29,7 +29,7 @@ class TestDeviceService extends GBDeviceService {
         // calling though to the service natively does not work with robolectric,
         // we have to use the ServiceController to do that
         service.onStartCommand(intent, Service.START_FLAG_REDELIVERY, (int) (Math.random() * 10000));
-//        super.invokeService(intent);
+        super.invokeService(intent);
     }
 
     @Override
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/impl/GBDeviceService.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/impl/GBDeviceService.java
@@ -19,10 +19,25 @@ import nodomain.freeyourgadget.gadgetbridge.model.MusicStateSpec;
 import nodomain.freeyourgadget.gadgetbridge.model.NotificationSpec;
 import nodomain.freeyourgadget.gadgetbridge.model.WeatherSpec;
 import nodomain.freeyourgadget.gadgetbridge.service.DeviceCommunicationService;
+import nodomain.freeyourgadget.gadgetbridge.util.LanguageUtils;
 
 public class GBDeviceService implements DeviceService {
     protected final Context mContext;
     private final Class<? extends Service> mServiceClass;
+    private final String[] transliterationExtras = new String[]{
+            EXTRA_NOTIFICATION_PHONENUMBER,
+            EXTRA_NOTIFICATION_SENDER,
+            EXTRA_NOTIFICATION_SUBJECT,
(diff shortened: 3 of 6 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
```diff
@@ -108,7 +109,13 @@ protected void onCreate(Bundle savedInstanceState) {
             @Override
             public void onClick(View v) {
                 NotificationSpec notificationSpec = new NotificationSpec();
-                String testString = editContent.getText().toString();
+                String testString = GetUserMessage();
+
+                if(LanguageUtils.transliterate())
```
cpfeiffer: AFAICS this *is* already transliterated, no?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
```diff
@@ -264,6 +271,18 @@ private void testNotification() {
         nManager.notify((int) System.currentTimeMillis(), ncomp.build());
     }
 
+    // Return user input from screen
+    private String GetUserMessage()
```
cpfeiffer: Nit: can you name it with a lower case first character, like getUserMessage()? Thanks :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/DebugActivity.java
```diff
@@ -180,9 +187,9 @@ public void onClick(View v) {
             @Override
             public void onClick(View v) {
                 MusicSpec musicSpec = new MusicSpec();
-                musicSpec.artist = editContent.getText().toString() + "(artist)";
-                musicSpec.album = editContent.getText().toString() + "(album)";
-                musicSpec.track = editContent.getText().toString() + "(track)";
+                musicSpec.artist = GetUserMessage() + "(artist)";
```
cpfeiffer: I guess a single call to getUserMessage() should be sufficient :-)

ivanovlev:
@cpfeiffer,

thx, fixed.

1. Left behind after extraction of method
2. agree
3. Sry, the habit from work

ivanovlev:
cpfeiffer,

thx, fixed.

left behind after extraction of method
agree
sry, the habit from work (c#)

ashimokawa:
@ivanovlev 
Of course the DeviceCommunicationService is used in the end. I don't understand the purpose of this PR

ivanovlev:
@ashimokawa

i see two way, how notification gets to device:
1. Intent from system, cathces by DeviceCommunicationService.**onStartCommand**. At the end calls interface methods DeviceSupport.**onNotification** \ **onSetMusicInfo** \ **onSetCallState** \ **onAddCalendarEvent**

2. Debug screen (DebugActivity), that generate NotificationSpec itself and call interface method **onNotification** independently.

Therefore, in the second case the transliteration has not been used

ashimokawa:
No you are mistaken. System events are caught by various receivers and they call the same method as in the debug screen.

ivanovlev:
Yep, onNotification, right? My mistake was that I have added a transliteration before call "onNotification"

ashimokawa:
If I am not totally blind you did it in DeviceCommunicationServer which gets always called in the end. A specific receiver would be SMSReceiver

danielegobbetti:
I just tested current master (without this PR) and by enabling transliteration the character æ is being transliterated to "ae", when sent from the debug screen.
I believe this pull request is not needed, if its purpose is making transliteration work from the debug screen.

ivanovlev:
Hmm, you all right. In this current master debug works normally. May be something was changed, since i seen bug.

Anyway, this PR makes a call transliteration more readable and understandable, it gathers all in one spot. It's good for others people. Then just need to rename the PR.

cpfeiffer:
AFAICS this PR (in its current state) moves the transliteration from `DeviceCommunicationService` to GBDeviceService, i.e. from the backend to the frontend, which is a good thing, IMHO.

ashimokawa:
@cpfeiffer 
Agreed, but I would squash it (because all the back and forth)

ivanovlev:
Okay, I'll try the evening

ivanovlev:
@ashimokawa It seems, i made it. Also i'm wrote a test - can you review it? I'm not sure, that i'm placed it in right place...

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/service/TestDeviceService.java
```diff
@@ -29,7 +29,7 @@ protected void invokeService(Intent intent) {
         // calling though to the service natively does not work with robolectric,
         // we have to use the ServiceController to do that
         service.onStartCommand(intent, Service.START_FLAG_REDELIVERY, (int) (Math.random() * 10000));
-//        super.invokeService(intent);
+        super.invokeService(intent);
```
cpfeiffer: Did you check that this works locally (without emulator or device)? See the comment above.

ivanovlev: I wanted to check transliterate support at whole. It is unlikely that someone will be able to break the method - it is written too simple;) But, give me a hour, i check and it...

cpfeiffer: See testFindDevice() for an example to test the whole thing.

ivanovlev:
@cpfeiffer 
I'm trying to understand, but it is impossible. How can i use mock class to catch call to real class inside 'service.invokeService(intent)';

cpfeiffer:
You can't mock static methods with Mockito, sorry. You can only create mock objects and verify what's being done with them in your test case.

PR #515, opened 2017-01-22
HPlus: Fix bug related to steps and heart rate

jpbarraca:
This PR will fix the display of todays steps and will ignore missing HeartRate values.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
@@ -391,6 +391,8 @@ class HPlusHandlerThread extends GBDeviceIoThread {
             sample.setRawHPlusHealthData(record.getRawData());
             sample.setProvider(provider);
 
+            provider.addGBActivitySample(sample);
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
```diff
@@ -391,14 +391,15 @@ public boolean processRealtimeStats(byte[] data) {
             sample.setRawHPlusHealthData(record.getRawData());
             sample.setProvider(provider);
 
+            provider.addGBActivitySample(sample);
+
             sample.setSteps(sample.getSteps() - prevRealTimeRecord.steps);
```
cpfeiffer: I think the steps will not be committed to the db when set *after* addGBActivitySample(). If that is intended, I suggest to add a short comment.

jpbarraca: They are set  in line 389, stored to the DB and then the sample steps are adjusted and sent to the intent. The reason is that we store the absolute number of steps (Important for the current day steps chart), but send the number of steps since the last update to the intent (for the real time data)

PR #517, opened 2017-01-24
HPlus: Improve display of new messages and phone calls

jpbarraca:
The Zeband supports a subset of the GB2312, with some changes (may still require further research):
* up to char 175 it seems to use ASCII/Extended ASCII
* chars 161-163, 169, and 176-247 give access to the GB2312 table and the device will process two bytes at once. It means that there are many accented and special characters. Cyrillic is missing...

This pull request improves text display by: 
* incorporating the Transliteration work recently introduced.
* adjusting the remaining latin characters as some are still present. Because the transliteration table is specific for this device, it was not added to the LanguageUtils (alternative solutions are welcome). The result is that Cyrillic and accented Latin characters will be (mostly) displayed properly.
*  Improving Message notification display: Notifications start by displaying a message icon and then the message itself. The message title will take the top row, while the body will take the bottom row of the display. The title is truncated (usually app name or contact name), while the body is displayed completely using multiple screens, but up to 127 screens.
* Improving Phonecall notification display: Call notifications start by displaying a call icon, the phone number, and then the caller name in the center of the screen.
* Reducing notification latency by using a single transaction for each notification type.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
@@ -44,7 +44,6 @@ import nodomain.freeyourgadget.gadgetbridge.service.btle.actions.SetDeviceStateA
 import nodomain.freeyourgadget.gadgetbridge.service.btle.profiles.deviceinfo.DeviceInfo;
 import nodomain.freeyourgadget.gadgetbridge.service.btle.profiles.deviceinfo.DeviceInfoProfile;
 import nodomain.freeyourgadget.gadgetbridge.util.GB;
-import nodomain.freeyourgadget.gadgetbridge.util.LanguageUtils;
 import nodomain.freeyourgadget.gadgetbridge.util.StringUtils;
 
 
@@ -677,10 +676,6 @@ public class HPlusSupport extends AbstractBTLEDeviceSupport {
             for (int i = 0; i < msg.length; i++)
                 msg[i] = ' ';
 
-            if(LanguageUtils.transliterate()) {
-                name = LanguageUtils.transliterate(name);
-            }
-
             byte[] nameBytes = encodeStringToDevice(name);
             for (int i = 0; i < nameBytes.length && i < (msg.length - 1); i++)
                 msg[i + 1] = nameBytes[i];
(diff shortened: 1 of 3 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -14,13 +14,16 @@
 import android.support.v4.content.LocalBroadcastManager;
 import android.widget.Toast;
 
+import org.apache.commons.lang3.ArrayUtils;
```
ashimokawa: Is that a new dependeny?

ashimokawa: Sorry just saw Carsten already had that for Mi Band 2 ..
Ohhh and I wrote stupid code because I wanted to prevent another dependency back then...

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -703,45 +703,44 @@ private void showText(String title, String body) {
...
-            for (int i = 0; i < msg.length; i++)
-                msg[i] = ' ';
+            String message = "";
 
-            msg[0] = HPlusConstants.CMD_ACTION_DISPLAY_TEXT;
+            if (title != null && title.length() > 0) {
 
-            String message = "";
+                if(LanguageUtils.transliterate()){
+                    title = LanguageUtils.transliterate(title);
```
ashimokawa: transliterate() is already called by generic code before reaching this code. So it should be removed here.

ivanovlev: Affirmative, it works before your code, see to the end of PR500 (DeviceCommunicationService) and PR 513 (GBDeviceService). But you can manually call it without check (even it off in settings)

jpbarraca: @ivanovlev  No need to call it manually. Will just keep the method to encode messages for the HPlus devices.

PR #523, opened 2017-01-26
HPlus: Support for Makibes F68

jpbarraca:
Adds support for the Makibes F68 Watch.
This is a subset of the HPlus devices with small changes. Most of the HPlus code is reused, with changes to consider both currently tested devices (Zeband and F68). There was some refactoring of existing code as my knowledge of the HPlus devices increased.

In both devices, if Continuous Heart Rate monitoring is not active, Activity Statistics will miss data. This will be addressed but I feel the code is ready for more testing as it enables these devices to connect.


ashimokawa:
very cool. I mergend this after tagging 0.17.3
Maybe you could use id 41 instead of 50 to make clear it is also part of the "HPlus group" ?

PR #532, opened 2017-01-30
HPlus: Working alarms and small cleanup

jpbarraca:
Properly enable alarms in HPlus devices
Small cleanup to simplify synchronisation process among the different devices


ashimokawa:
Thanks! Sorry for the delay

PR #541, opened 2017-02-04
Transliteration: capitalized just the first letter in the replacement

ivanovlev:
Small fix to upper case scenario:

old variant:    "Щавелевый суп" > "SHHavelevyjj sup"
fixed variant: "Щавелевый суп" > "Shhavelevyjj sup"


ivanovlev:
I thought about it, but it's not critical changes. May be later

PR #544, opened 2017-02-06
Background javascript

chriskoups:
sorry, thats a wrong pull request! it was created by accident and I don't know how to remove it!!!!!


danielegobbetti:
You can just hit the "close" button and it will be closed without merging (so it will have no effect on the main codebase). I did it for you.

PR #545, opened 2017-02-06
Target step on the "About You"

chriskoups:
Move target step from miband options to about you option.
ATTENTION
App crashes when I tried to access the miband settings..! I don't know all the bits of the code so, if you don't know how to fix the crash do not merge!!!!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ###Changelog
 
+####Version next
+* Better integration with android music players
+
 ####Version 0.17.3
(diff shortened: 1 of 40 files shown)


danielegobbetti:
You added unrelated commits taken from another branch and did two commits (update preferences.xml and miband...) Where one would suffice.
The commit message should allow other people browsing the log to understand what's going on, so "update file xyz" is not a good commit message. See http://chris.beams.io/posts/git-commit/ for a better explanation.

For these reasons I am closing this, please open a new pull request once you fixed these two issues.

PR #548, opened 2017-02-06
Update HPlusSupport.java, Swap day and month

chriskoups:
I have swap day and month because it is more logical to ready first the day and then the month. Plus the day would be with bold and will be the first thing that you will check.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
@@ -199,8 +199,8 @@ public class HPlusSupport extends AbstractBTLEDeviceSupport {
                 HPlusConstants.CMD_SET_DATE,
                 (byte) ((year / 256) & 0xff),
                 (byte) (year % 256),
-                (byte) (month + 1),
-                (byte) (day)
+                (byte) (day),
+                (byte) (month + 1)
 
         });
         return this;


jpbarraca:
Hi @chriskoups.  You cannot do that  :) 
The device has no means to know the order of the fields in the array. So, the message structure is fixed and preset in the firmware. This applies to all messages sent and received.

danielegobbetti:
Even if the change was correct, please remember that commit messages *must* be meaningful and not simply "update file xyz" otherwise they will not be merged.

cpfeiffer:
@chriskoups Did you actually test your change and did it work? If it it did appear to work, I guess it was pure chance because the current day is smaller than 12 (highest month).

chriskoups:
I try it and I think I change the day through my phone and everything was ok! But as @jpbarraca said, it something fixed. And when it saves the data are saved messed up. Unfortunately, I don't have the time now to check out all the date path so for now i can do something. Don't forget that is an electronic watch and if you wrote potato  there it will show potato :P

cpfeiffer:
Well, feel free to create your custom Potatobridge :-)

chriskoups:
Hahaha... During weekends... If you know how to swap those two feel free to help me :)

cpfeiffer:
Sorry, can't help you with that, I don't even have such a device. But if you really want to have this, invest some time in understanding the device and its Bluetooth API. You would look for a way to change the date *format*, i.e. the way it is displayed, not the way how the date is sent to the device.

jpbarraca:
@chriskoups I found how to configure the most standard things, but there may be other aspects that could be dynamically modified. However, I have no idea how to do it.

One thing that puzzles me is how the message IDs were selected and what other messages may be available.

chriskoups:
I will have a look tomorrow and during the weekend..... If you found any documentation for the device please send it to me (I know that most probably you did not)... I agree that they are some more important things to be fixed.....!
I found out that the HR is not set to all day record because when I close my Bluetooth is stop recording! I thing I "fixed" by removing the disable command but I did not had time to fail proof it... Also, the symbol on the device change from just heart to heart and a cross on top, after i removed the command!

jpbarraca:
@chriskoups I have no documentation.
The Hall Day HR feature will make the device issue a message every time there is movement and will measure the HR periodically. Every 5 minutes there is another message that is sent, which may contain the average HR of the previous 5m slot. Moreover, you can get the log of the last 24h in slots of 10m.

If you enable All Day HR and stop the application, the real time messages will be lost and you will only get the 10m averages.

chriskoups:
Ok....... I really want to understand the whole code but is uncommented and I got confused....
I am sure that in my application i don't get the 10mins averages because the graph is hitting 0 (actually it goes less than 0)... If you can guide me in which part of the code I may find those settings I may play around them....! For now I try to introduce REM sleep in the graph... 

P.S. Today it start measuring deep sleep between 13:00-!5:00 while I was in the office and I was typing!!! :D

jpbarraca:
The chart will show 0 if All Day HR is off. Basically because the smart watch will not be measuring any HR. Steps counts, calories and distance will be provided but only steps will be used.

Sleep detection seems to be somewhat inaccurate with these devices. Currently the sleep periods are shown as provided by the smart watch. The only post-processing done is to insert "virtual samples" near the boundaries of the overlays so that the accuracy is increased. Ideas of how to improve this are also much welcome.

In order to test command you can either add them directly in GadgetBridge (HPlusSupport.java) or use an external tool such as https://github.com/nccgroup/BLE-Replay

Issue #536, opened 2017-01-31
Pebble music control PowerAMP track data


Avamander:
Using PowerAMP to play my music the Pebble music control app shows nothing where it should show the title, length and artist.

Pebble Time 4.3 (snowy_dvt)
Android 4.4.4 (LG D605)
GadgetBridge 0.7.12

Avamander:
http://forum.powerampapp.com/index.php?/topic/5511-can-i-send-current-song-info-to-a-web-server/&do=findComment&comment=21114

"Poweramp sends current track info via own API and via standard Android API, so you can listen to one of those (with Broadcast Receiver) and re-send to whatever target you need (e.g. http server). Check our SDK for example app."

https://github.com/maxmpz/powerampapi/blob/master/poweramp_api_lib/src/com/maxmpz/poweramp/player/PowerampAPI.java

danielegobbetti:
The issue is (possibly) fixed in master, as now we listen for more actions by the standard android API. I didn't test with poweramp as I don't have google play store on my devices.

Avamander:
@danielegobbetti http://powerampapp.com/download-poweramp/

danielegobbetti:
Installed on a spare device (API level 19) and seemed to work well except for the position updates that are not reported (i.e. song name and artist and total duration are shown, but the progress bar remains stuck at 00:00).

Avamander:
Tested the latest master branch on my LG D605, the artist and song title are successfully displayed on my Pebble Time, nice work!

Avamander:
https://github.com/Freeyourgadget/Gadgetbridge/pull/550

danielegobbetti:
@Avamander since the update hit fdroid can you please confirm the issue is resolved and, if yes, close.

Avamander:
Not resolved fully as the progress is not being updated unless an intent is received. 
a) Track progress is stuck
b) When disconnected the status is lost and not updated

danielegobbetti:
That part is actually solved on android version >= 5 where the media notifications are part of the operating system provided features.

Avamander:
But that still means it's not fully solved though.

PR #554, opened 2017-02-14
Extract music shuffle and repeat states and set the song progress to auto-update.

Avamander:
I don't see it used anywhere, but it might be useful to have that data too. Though as Pebbles do nothing with the data I can't verify it being 100% correct. (Squash merge is recommended again as it would keep the history cleaner)

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/MusicPlaybackReceiver.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/MusicPlaybackReceiver.java
@@ -4,6 +4,7 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
+import android.os.Handler;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


Avamander:
Though I still lack the knowledge to make the progress bar smoothly update. It would be really lovely if someone pointed me in the right direction.

Avamander:
Okay, I made a small Timer to update the seconds continuously. I did what I could.

ashimokawa:
@Avamander 
I dont understand why you update the progress. The pebble does this automatically.

Avamander:
I'm using Android 4.4 and most certainly the progress is not updated.

ashimokawa:
@Avamander
As long as the pebble knows the song is playing it progresses automatically, has nothing to do with the Android version.

Avamander:
Well the progress bar on my Pebble Time says otherwise. It's literally not updating unless a media intent is received.

ashimokawa:
@Avamander 
Then I am 100% sure the pebble is not in "playing" state in your case - it does update when it is playing. On the Pebble Time.

Avamander:
@ashimokawa What state can it be in if it doesn't show the "pause" icon and it easily switches between playing and not playing icons? It's playing, the Pebble knows it, I know it, the app knows it but the progress is not updating. See https://github.com/Freeyourgadget/Gadgetbridge/issues/536

ashimokawa:
@Avamander 
Do other players work? Trust me if everything is done correctly from Gadgetbridge the pebble music app would update the position automatically. There has to another problem.

Avamander:
I also tried Google Play Music, using the latest master without my continuous update code, no it does not update the position automatically. As far as I understood danielegobbetti it's a known problem.

Though I think I found a bug with seeking while paused, need to fix that.

danielegobbetti:
I'm from mobile and didn't check neither original code nor the PR, but perhaps a missing/wrong playing rate could be the culprit?

Avamander:
If that is the case I'll just set the right play speed and remove the unnecessary code.

Avamander:
Did that and it should be fine now. Thanks for the information about playRate.

ashimokawa:
@Avamander

Please remove all formatting changes in MusicStateSpec.java

ashimokawa:
@Avamander 
Thanks!

Avamander:
Always happy to help.

PR #568, opened 2017-02-28
Added Hebrew transliteration and test

yarons:
Please excuse me if my test case is wrong, I'd love to change it if needed.

Thank you!


cpfeiffer:
Thanks for your contribution! AFAICS, the result should be "bdykh 'bryth", right? Can you fix the testcase and push your branch again?

yarons:
Changed it in my branch, hoping that's fine, do you want me to add another transliteration?

There is a very good system here, the funny thing is that we don't need the Chinese conversion because Mi Band already has Chinese support.

yarons:
I'm not fully aware of the situation with values-pt_BR but I can add this again to the latest branch if it helps.

PR #581, opened 2017-03-03
display call

albertsal:
i change row 168 and 169

in this way it will be shown before the name and later the number

from

Intent intent = createIntent().setAction(ACTION_CALLSTATE)
.putExtra(EXTRA_CALL_PHONENUMBER, callSpec.number)
.putExtra(EXTRA_CALL_DISPLAYNAME, callSpec.name)
.putExtra(EXTRA_CALL_COMMAND, callSpec.command);
invokeService(intent);

to

Intent intent = createIntent().setAction(ACTION_CALLSTATE)
.putExtra(EXTRA_CALL_DISPLAYNAME, callSpec.name)
.putExtra(EXTRA_CALL_PHONENUMBER, callSpec.number)
.putExtra(EXTRA_CALL_COMMAND, callSpec.command);
invokeService(intent);


ashimokawa:
Changing the order here does not change anything on the actual device. Your change does nothing.

Also, the PR is totally broken and contains a lot of unrelated commits.

cpfeiffer:
@albertsal If you wish to change the order you should have a look at PebbleProtocol.encodeSetCallState().

Not sure whether this is possible at all though. 

Before you submit something, please verify that your change actually works, and, if it changes existing behavior, you should think about whether you should change that for *all* users or whether there should be a configuration option.

PR #590, opened 2017-03-05
Fixed Hebrew transliteration test

yarons:
There you go, sorry.


yarons:
Yes the same, may I should escape the ' char?

cpfeiffer:
I fixed the testcases, see 0b45fe63f00860e26b818ac399095a65121cf1b1 and check if the expected transliteration is correct. If it's incorrect, the transliteration mapping is wrong.

PR #618, opened 2017-03-26
Addon hplus + import export SharedPreferences

albertsal:
I added the notification of low battery HPlus ... specifically, I used the HPlusConstants.DATA_STATS event on hplusSupport

I also included the status of the device in the list of GBDeviceAdapterv2

and 

i add function for import and export SharedPreferences setting

when export or import db execute export or import SharedPreferences

for blacklist i preload HashSet


jpbarraca:
These seem to be two different PR (shared preferences and hplus stuff)

albertsal:
@jpbarraca 
I created another branch for HPlus but I think the github app has included as well the other commit

Issue #442, opened 2016-12-01
Sleep graphs are inconsistent and have gaps(?)


6arms1leg:
There are three things I noticed about the sleep activity graphs which I think are wrong(?):

* "Light Sleep" and "Deep Sleep" seem to be inverted: The sleep bars with a higher score ("more activity") are interpreted as "Deep Sleep", whereas sleep bars with low scores ("lower activity") are shown as "Light Sleep". Now, I'm not an expert on sleep but intuitively I would expect to see exactly the opposite. Or is it in fact so that a person is more active in deep sleep than in light sleep phases?
* The data in the sleep pie chart doesn't match the other graphs. E.g.: While the sleep pie chart shows about 10% "Deep Sleep", the sleep bar chart shows that almost all sleep is "Deep Sleep".
* The sleep activity bar graph shows huge gaps between the activity bars, often with no data for about 30min or more. During the day (green bars) there are no gaps at all.

I observed the above with my old Mi Band 1.

cpfeiffer:
I guess we should make the labels of deep and light sleep user-configurable (SCNR ;-)).

We had those labels inverted previously, but some people reported that they were wrong and Mi Fit would report them the other way round, so we toggled the meaning, which is still how it is today.

We had a few problems when we ported to a major new version of the charting library, so it might be that there are still some regressions.

It may very well be that there *is no* activity data for a while though. If you want to share a gadgetbridge log from a day where this happens (together with a screenshot of the charts view), please do. I could check if the band didn't provide any data or if there's something wrong.

6arms1leg:
As always, many thanks for your support!

Here is a Gadgetbridge log and the corresponding screenshots of the sleep screen in Gadgetbridge (one with the bar graph zoomed to the whole sleep part only and one zoomed to a ~30min gap):
[GB_Sleep_Issue-442.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/629517/GB_Sleep_Issue-442.zip)

The Mi Band 1's sleep detection is awful, as I went to bed at around 02:00am, but I guess that's a problem with the device itself, not Gadgetbridge.

> It may very well be that there is no activity data for a while though.

Ah OK, I checked the log and the Mi Band 1 indeed reports an "intensity" of 0.0 for long periods of time that match the gaps of the bar graph in Gadgetbridge. So Gadgetbridge just displays them as gaps in the bar graph but actually takes them into account for the sleep pie chart.
[...]

PR #636, opened 2017-04-01
Swap deep and light sleep colours in charts

walkjivefly:
This partially "fixes" issue #442. The effect of the change is
purely cosmetic: deep sleep is (once again) dark blue, light sleep
is (once again) light blue.


walkjivefly:
I'm confused. If I click "See review" above nothing happens. Is there any more to see than your comment about changing the constants, not just the fragment? 

I believe I've done it the way you ask. In colours.xml there are already
```
    <color name="chart_deep_sleep_light" type="color">#0071b7</color>
    <color name="chart_deep_sleep_dark" type="color">#4c5aff</color>
    <color name="chart_light_sleep_light" type="color">#46acea</color>
    <color name="chart_light_sleep_dark" type="color">#b6bfff</color>
```

Which look [correct](http://imgur.com/a/AjsGr) to me. But the fragment was pseudocode setting 
```
AK_DEEP_SLEEP_COLOR = R.attr.chart_light_sleep.data
AK_LIGHT_SLEEP_COLOR = R.attr.chart_deep_sleep.data
```

So I just switched them back here. Apologies if I'm missing something else.

cpfeiffer:
You're totally right, I misread the diff. Thanks!

PR #638, opened 2017-04-01
HPlus: added low battery information

albertsal:
I added the notification of low battery HPlus ... specifically, I used
the HPlusConstants.DATA_STATS event on hplusSupport

I also included the status of the device in the list of
gbdevice -> getDeviceInfos

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/impl/GBDevice.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/impl/GBDevice.java
@@ -78,6 +78,15 @@ public class GBDevice implements Parcelable {
     private String mBusyTask;
     private List<ItemWithDetails> mDeviceInfos;
 
+    private int mStep = 0;
+    private int mCalory = 0;
+    private int mHeart = 0;
+    private int mDistance = 0;
+    private static final String DEVINFO_STEP = GBApplication.getContext().getString(R.string.chart_steps) + ": ";
+    private static final String DEVINFO_DISTANCE = GBApplication.getContext().getString(R.string.distance) + ": ";
+    private static final String DEVINFO_CALORY = GBApplication.getContext().getString(R.string.calories) + ": ";
+    private static final String DEVINFO_HEART = "HR: ";
+
     public GBDevice(String address, String name, DeviceType deviceType) {
         this(address, null, name, deviceType);
     }
@@ -106,6 +115,11 @@ public class GBDevice implements Parcelable {
         mBusyTask = in.readString();
         mDeviceInfos = in.readArrayList(getClass().getClassLoader());
 
+        mStep = in.readInt();
+        mDistance = in.readInt();
+        mCalory = in.readInt();
+        mHeart = in.readInt();
+
         validate();
     }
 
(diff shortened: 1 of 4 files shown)


albertsal:
hi @jpbarraca  i have correct pr

jpbarraca:
I don't feel confortable with such specific fields being added to GBDevice as I think a more generic approach could be followed. As an example, through a collection of generic slots that each device handler could use to provide device specific information.

@cpfeiffer : can you comment on this?

albertsal:
I saw that it lacks a global summary of activities so I thought I'd add it.
This usually has to be on the first screen just have not found a valid space

jpbarraca:
@albertsal I actually can agree with the concept but I think the implementation can be different.

albertsal:
ok I accept suggestions

jpbarraca:
Different devices will have different metrics to be presented. 
Adding specific properties to GBDevice such as Calories, Steps, etc... is only valid for HPlus devices and GBDevice is a more generic class.

Therefore, IMHO your PR makes sense if the Device panel shows data from a collection of items with a known format (interface). This interface could have a title, value (and a level?). The level (which I think is clearly optional) could be used to display some values in a different color if they represent a critical or warn situation.

Other approach would be for GBDevice to be abstract and each device implements its interface in a device specific manner.

However this requires a discussion about the GB user interface, which is already in progress.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/impl/GBDevice.java
```diff
@@ -78,6 +78,15 @@ public GBDevice createFromParcel(Parcel source) {
     private String mBusyTask;
     private List<ItemWithDetails> mDeviceInfos;
 
+    private int mStep = 0;
```
cpfeiffer: @jpbarraca is right -- this is device specific and should not be there. However, there is already support for adding "any kind" of additional information to GBDevice. See method GBDevice#addDeviceInfo(). Can you try to use that?

albertsal: ok i try

albertsal:
hi @jpbarraca
I changed all the logic now in use #addDeviceInfo as suggested by
@cpfeiffer
I think now it's ok

cpfeiffer:
Yes, this looks much better, thanks! From my point of view only the method name should be improved (Extra_info() => processExtraInfo() or something like that.

Thanks again :+1:

cpfeiffer:
BTW, you don't *have* to revert commits. You can also just add your changes in another commit and we will squash all of them together before merging.

albertsal:
you're right, but I thought it was more properly the revert, since I have changed all

cpfeiffer:
So from my point of view, we can merge it. @jpbarraca, what do you think? Can we merge this or do we need a better mechanism for displaying the information?

jpbarraca:
I think it is ok.

cpfeiffer:
Uh. I was just about to merge it. Now you added another commit and we need to take another round of review. How about resetting the branch to the previous commit, force-pushing it and opening another branch for the lastChange thing?

albertsal:
I'm sorry I thought you were in time to add this piece ... I take it out and add it later

PR #641, opened 2017-04-04
Piechart:graphic optimization pie sleep

albertsal:
on my galaxy s6 I noticed that the pie chart to sleep is difficult to
read because the label are above the values, I thought it was better to
use a legend and change the color of the values ...

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractWeekChartFragment.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractWeekChartFragment.java
@@ -141,6 +141,7 @@ public abstract class AbstractWeekChartFragment extends AbstractChartFragment {
         }
 
         set.setValueTextColor(Color.WHITE);
+        set.setValueTextSize(15f);
         set.setValueFormatter(getPieValueFormatter());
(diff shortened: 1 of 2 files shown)


cpfeiffer:
Thanks for working on this, however I did not merge this, because the legend would mostly be duplicated among the two charts. Both charts use the same colors, so no need to have two legends with the same colors and labels.

I did try to improve the labelling in the pie chart though, let me know if you like it, or feel free to improve it.

albertsal:
Ok tk

albertsal:
screenshot_20170405-081924

I am attaching a screenshot of the previous version in order to see the problem

albertsal:
screenshot_20170405-093524


on this current version ....
how do you see the character of the values is too small to any graphics ... I suggest you zoom in all the graphic characters and maybe to use white color for the pie charts

danielegobbetti:
Update your fork to the latest master! I see from your screenshots that you are probably not aligned with the latest changes.

Just to be clear: this is not related to the pull request content, but you should ensure you keep your fork up to date to avoid conflicts.

albertsal:
Now I put the last update, and the situation is this

screenshot_20170405-193320
screenshot_20170405-193334
screenshot_20170405-193341

PR #644, opened 2017-04-07
Hplus:last charge

albertsal:
I created a function to determine how much time has passed since last
charge


albertsal:
I am having swings more than 5%
I close pr so do far more tests

PR #645, opened 2017-04-07
Font size bar chart

albertsal:
small fix that improves the readability of the values on bar charts


cpfeiffer:
Hardcoding BLACK makes it pretty much unreadable in the dark theme. @danielegobbetti is working on the styles, so maybe we will get properly styled charts at some point. At the moment, the hardcoded GRAY works for both themes.

Changing the font size is OK though.

albertsal:
ok tk

cpfeiffer:
Thanks, Alberto!

albertsal:
;)

PR #646, opened 2017-04-07
Device info display

albertsal:
small fix handy when are shown more than 3 info


danielegobbetti:
Thanks for the contribution! Hopefully we will not have too much line of text :-)

albertsal:
;)

PR #647, opened 2017-04-07
Label piechart

albertsal:
small fix that removes pies labels.
I noticed that the labels have a white font and you can see if it moves
the graph


cpfeiffer:
I don't quite understand -- do you have a screenshot of what's wrong?

albertsal:
image

I changed a bit the contrast so you can see better the problem

cpfeiffer:
That's how it looks with your patch, but what was wrong with showing the label?

danielegobbetti:
On the light theme the first line "Xh YYmin" is visible, the second line "Light sleep"/"Deep sleep" is white on white.

danielegobbetti:
@albertsal your PR pointed me in the right direction to catch the bug, but I fixed it by showing the labels also on the light theme instead of hiding it :-)

albertsal:
Yes @danielegobbetti the label is white and background is white.
But you rotate the chart the label go on  background blu of the chart and read partial label...

I think it is not necessary that label if the legend down there

albertsal:
Ok @danielegobbetti I read just now your last post

PR #652, opened 2017-04-13
BATTERY THRESHOLD PERCENT

albertsal:
I added the option to set the battery threshold by tapping the battery
icon, I made sure each device has its own threshold

HPLUS: I put a small correction to show the low battery notification

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/adapter/GBDeviceAdapterv2.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/adapter/GBDeviceAdapterv2.java
@@ -22,15 +22,19 @@ import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.support.design.widget.Snackbar;
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v7.widget.CardView;
 import android.support.v7.widget.RecyclerView;
+import android.text.InputType;
 import android.transition.TransitionManager;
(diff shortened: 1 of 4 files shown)


jpbarraca:
I would prefer this to be in the HPlus settings panel.
The objective is to keep all settings in the same place instead of scattering configuration points across the app.

albertsal:
Initially I did this, but I noticed that each device can have a different threshold. For example i have 2 devices the first device should be set between 15% and 20% because below this threshold goes off, for the second device it is ok 10%

jpbarraca:
Ideally we would have completely different settings for different devices.
@cpfeiffer : how is this handled with Mi Band and Pebble?

albertsal:
I have 2 device for hplus ... zeband and s200

cpfeiffer:
Mi Bands provide the battery level on request. Additionally they notify low battery state on their own. I think we have a hardcoded threshold somewhere that colours the displayed battery level in red.

For Pebble, the whole battery handling is WIP, AFAIK.

IMHO this need not be user configurable. We could hardcode device specific thresholds though (need feedback on the different devices though, maybe with a wiki page?)

albertsal:
@cpfeiffer ok is clear... so it is possible that some devices HPlus do not have  notification ... @jpbarraca your devices have notified?

cpfeiffer:
@albertsal Yes, that's certainly possible. I do not have any HPlus device yet, so I can't say if *any* HPlus device provide battery notifications.

albertsal:
@cpfeiffer  Ok i could just put a single global setting for all devices or put it on hplus only ... but if you think we'll soon think of this change we can close the pr

PR #653, opened 2017-04-13
HPLUS:last charge

albertsal:
I created the logic to determine how much time has passed since the last
charge.
The device has many oscillations on the current charge so i stabilized
the value by sampling the values collected

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
@@ -363,13 +363,14 @@ class HPlusHandlerThread extends GBDeviceIoThread {
         //Skip when measuring heart rate
         //Calories and Distance are updated and these values will be lost.
         //Because a message with a valid Heart Rate will be provided, this loss very limited
-        if(record.heartRate == ActivityKind.TYPE_NOT_MEASURED) {
+        getDevice().setFirmwareVersion2(null);
+      /*  if(record.heartRate == ActivityKind.TYPE_NOT_MEASURED) {
             getDevice().setFirmwareVersion2("---");
             getDevice().sendDeviceUpdateIntent(getContext());
         }else {
             getDevice().setFirmwareVersion2("" + record.heartRate);
             getDevice().sendDeviceUpdateIntent(getContext());
-        }
+        }*/
 
         try (DBHandler dbHandler = GBApplication.acquireDB()) {
             HPlusHealthSampleProvider provider = new HPlusHealthSampleProvider(getDevice(), dbHandler.getDaoSession());
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
@@ -900,6 +900,12 @@ public class HPlusSupport extends AbstractBTLEDeviceSupport {
         }
         short level = bNormalize.getLevelNormalize((int) data,batteryCmd);
 
+        LOG.debug("Count:" + String.valueOf(bNormalize.getCounter()) +
+                " Mid:" + String.valueOf((int) level) +
+                " Band:" + String.valueOf((int) data) +
+                " Fix:" + String.valueOf(batteryCmd.level) +
+                " Time:" + String.valueOf(batteryCmd.lastChargeTime.getTimeInMillis()));
+
         if (batteryCmd.level != level) {
             SharedPreferences.Editor editor = settings.edit();
             if (level > batteryCmd.level) {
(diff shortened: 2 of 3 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/BatteryNormalize.java
```diff
@@ -0,0 +1,44 @@
...
+    private static int INTERVAL_SECOND_CHECK  = 30;
+    private static int MAX_COUNT_CHECK = 6;
+
+    public void BatteryNormalize(int interval,int numerberCheck){
+        INTERVAL_SECOND_CHECK=interval;
+        MAX_COUNT_CHECK=numerberCheck;
+    }
+
+    //use this function if the battery level of the device oscillates
+    public short getLevelNormalize(int data, GBDeviceEventBatteryInfo cmd) {
```
cpfeiffer: I'm not sure I completely understand how this class works, but it looks like you're basically just skipping some values. But how do you know that the value you're actually using is not one of the "wrong" values? Or do you just want to use *fewer* values to keep them from oscillating?

Did you think about using the median value instead?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
```diff
@@ -363,13 +363,14 @@ public boolean processRealtimeStats(byte[] data) {
         //Skip when measuring heart rate
         //Calories and Distance are updated and these values will be lost.
         //Because a message with a valid Heart Rate will be provided, this loss very limited
-        if(record.heartRate == ActivityKind.TYPE_NOT_MEASURED) {
+        getDevice().setFirmwareVersion2(null);
+      /*  if(record.heartRate == ActivityKind.TYPE_NOT_MEASURED) {
```
cpfeiffer: Did you change that on purpose for this PR?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -870,8 +877,8 @@ private void  processExtraInfo (byte[] data) {
             if (record.calories > 0) {
                 info += DEVINFO_CALORY + String.valueOf(record.calories) + "   ";
             }
-            if (record.heartRate > 0) {
-                info += DEVINFO_HEART + String.valueOf(record.heartRate) + "   ";
+            if (record.heartRate > 0 && record.heartRate != ActivityKind.TYPE_NOT_MEASURED) {
+                info += "\n" + DEVINFO_HEART + String.valueOf(record.heartRate);
```
cpfeiffer: Same as above, this doesn't appear to be related to the battery.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
```diff
@@ -882,11 +889,33 @@ private void  processExtraInfo (byte[] data) {
         }
     }
 
-    private void handleBatteryInfo(byte data) {
-            if (batteryCmd.level != (short) data) {
-                batteryCmd.level = (short) data;
-                handleGBDeviceEvent(batteryCmd);
+    private String getLastCharge(byte data) {
```
cpfeiffer: The method name implies a simple conversion/calculation method. Could you rename it so that one becomes aware of its side effects (e.g. `determineLastChargeTimeAndNotify()`). A bit long, yes, but that's what you get when doing multiple things in a single method :-]

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/BatteryNormalize.java
albertsal: I try to explain it without log ... it often happens that the battery level drops for a few seconds or minutes and then goes up, the difference ranges between 2 and 20 percent so I thought I would spot 6 values ​​with a 30 second interval where recovery Only the most other value at the end of the 6 controls I return the highest value that is the correct one, so I understand if the battery has actually been discharged or charged or you have temporarily lowered the battery level

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusHandlerThread.java
albertsal: Not really but to get a clearer idea of ​​what I did, I preferred to move the whole logic of information to hplussupport

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
albertsal: Same as above

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
albertsal: Ok I change to the next commit

PR #672, opened 2017-04-28
Refactor notification management

danielegobbetti:
- centralize the logic for skipping over unwanted notifications
- use *Compat methods wherever possible
- use unique and persistent ID (update notifications)
- switch to using BigText style by default (since we can now update existing notifications)
- for Pebble: delete and reinsert notification as updating is not possible

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/NotificationListener.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/NotificationListener.java
@@ -31,6 +31,7 @@ import android.content.pm.PackageManager;
 import android.media.MediaMetadata;
 import android.media.session.PlaybackState;
 import android.os.Bundle;
+import android.os.Parcelable;
 import android.os.PowerManager;
 import android.os.RemoteException;
 import android.service.notification.NotificationListenerService;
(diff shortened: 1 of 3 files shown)


danielegobbetti:
This PR needs to be tested from other devices before merging. @jpbarraca can you please test HPlus devices?

On Pebble all the functionalities are working as intended (including: music control, replying to actionable notifications, muting, dismissing etc.).

It might be possible that the same workaround of deleting + resending the notification is needed also on other devices, in this case this can be done in the device Support class (see PebbleSupport changes in this PR).

Add commit to the branch in case further changes are needed (this PR is from a local branch).

danielegobbetti:
After further testing there are some drawbacks vs current master:

- conversations sends updates with the same notification id, but does not resend the whole notification content (append mode)
- signal does the same, but is sending the whole notification content (replace mode)
- signal uses the same notification id for each contact (or group), but if there are multiple unread notification ids it will resend them upon checking one of the "threads"

I am at this point unsure if it's worthwhile to proceed with the PR or if it's better to start from scratch.

danielegobbetti:
Conversations issues seem fixed now.

danielegobbetti:
The signal issue mentioned above:
> signal uses the same notification id for each contact (or group), but if there are multiple unread notification ids it will resend them upon checking one of the "threads"

Is present also in master, so there is **no known drawback** in this PR. Waiting for test on Hplus devices @jpbarraca .

jpbarraca:
Under test. Will report in the end of the day.

jpbarraca:
Don't know if this is strictly related to this PR but notifications are not caught by GB. 
The `NotificationListnener.onNotificationPosted` is never called in my phone.
Notifications for Incoming Calls work as expected.

danielegobbetti:
No that is unfortunately unrelated. You can try looking for "Collector" entries in the log. That service is supposed to revive the notification listener when needed, but we're experiencing mixed results with it. Seems to work fairly well with aosp, not so much on cm/lineageos.

danielegobbetti:
Just saw your edit. Cool, thanks!

jpbarraca:
The method was never called, but i'm not sure what was the state of the notification listener. 
I rebooted the phone and it has been working flawlessly since that moment.

If this happens again I will try to determine if the listener service is active and debug the issue.
This is an Android 7.0, with Huawei EMUI 5.0

danielegobbetti:
You should see lines mentioning the collector monitor service.

danielegobbetti:
Changed the PR title as it is tested and could be merged.

PR #680, opened 2017-05-08
App installation screen improvements

rosenpin:
Improvements now include:

- Users can now choose multiple files to install at once
- Progress bar and an appropriate message when the app is successfully/unsuccessfully installed
- Replaced RelativeLayout with LinearLayout where it seems to be more appropriate (activity_appinstaller.xml)

preview

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -66,7 +66,7 @@
         <activity
             android:name=".activities.FwAppInstallerActivity"
             android:label="@string/title_activity_fw_app_insaller"
-            android:parentActivityName=".activities.ControlCenterv2">
+            android:theme="@style/Theme.AppCompat.Light.NoActionBar">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -65,8 +65,7 @@
             android:parentActivityName=".activities.SettingsActivity" />
         <activity
             android:name=".activities.FwAppInstallerActivity"
-            android:label="@string/title_activity_fw_app_insaller"
-            android:theme="@style/Theme.AppCompat.Light.NoActionBar">
+            android:label="@string/title_activity_fw_app_insaller">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
@@ -24,7 +24,9 @@ import android.content.IntentFilter;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.v4.app.NavUtils;
+import android.support.v4.content.ContextCompat;
 import android.support.v4.content.LocalBroadcastManager;
+import android.support.v7.widget.Toolbar;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.Button;
@@ -55,17 +57,15 @@ public class FwAppInstallerActivity extends GBActivity implements InstallActivit
 
     private static final Logger LOG = LoggerFactory.getLogger(FwAppInstallerActivity.class);
     private static final String ITEM_DETAILS = "details";
-
+    private final List<ItemWithDetails> mItems = new ArrayList<>();
     private TextView fwAppInstallTextView;
     private Button installButton;
     private Uri uri;
     private GBDevice device;
     private InstallHandler installHandler;
     private boolean mayConnect;
-
     private ProgressBar mProgressBar;
     private ListView itemListView;
-    private final List<ItemWithDetails> mItems = new ArrayList<>();
     private ItemWithDetailsAdapter mItemAdapter;
 
     private ListView detailsListView;
@@ -128,6 +128,12 @@ public class FwAppInstallerActivity extends GBActivity implements InstallActivit
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        if (GBApplication.isDarkThemeEnabled()) {
+            setTheme(R.style.GadgetbridgeThemeDark_NoActionBar);
+        } else {
+            setTheme(R.style.GadgetbridgeTheme_NoActionBar);
+        }
+
         setContentView(R.layout.activity_appinstaller);
 
         GBDevice dev = getIntent().getParcelableExtra(GBDevice.EXTRA_DEVICE);
@@ -172,6 +178,13 @@ public class FwAppInstallerActivity extends GBActivity implements InstallActivit
         if (uri == null) { //for "share" intent
             uri = getIntent().getParcelableExtra(Intent.EXTRA_STREAM);
         }
+
+        int appsCount = getIntent().getIntExtra("APPS_COUNT", 1);
+        int currentAppIndex = getIntent().getIntExtra("APP_INDEX", 1);
+        Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
+        toolbar.setTitleTextColor(ContextCompat.getColor(getApplicationContext(), android.R.color.white));
+        toolbar.setTitle("(" + currentAppIndex + "/" + appsCount + ")");
+
         installHandler = findInstallHandlerFor(uri);
         if (installHandler == null) {
             setInfoText(getString(R.string.installer_activity_unable_to_find_handler));
@@ -220,13 +233,13 @@ public class FwAppInstallerActivity extends GBActivity implements InstallActivit
     }
 
     @Override
-    public void setInfoText(String text) {
-        fwAppInstallTextView.setText(text);
+    public CharSequence getInfoText() {
+        return fwAppInstallTextView.getText();
     }
 
     @Override
-    public CharSequence getInfoText() {
-        return fwAppInstallTextView.getText();
+    public void setInfoText(String text) {
+        fwAppInstallTextView.setText(text);
     }
 
     @Override
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
@@ -77,6 +77,7 @@ public class FwAppInstallerActivity extends GBActivity implements InstallActivit
     private final BroadcastReceiver installationResultReceiver = new BroadcastReceiver() {
         @Override
         public void onReceive(Context context, Intent intent) {
+            //Receive installation fail/success intent
             if (mProgressBar != null)
                 mProgressBar.setVisibility(View.INVISIBLE);
             if (installButton != null) {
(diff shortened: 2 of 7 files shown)


danielegobbetti:
This is really interesting, thanks!

Especially wrt the layout changes, this should be carefully tested also with firmware installation and for other devices as well.

app/src/main/AndroidManifest.xml
```diff
@@ -65,8 +65,7 @@
             android:parentActivityName=".activities.SettingsActivity" />
         <activity
             android:name=".activities.FwAppInstallerActivity"
-            android:label="@string/title_activity_fw_app_insaller"
-            android:parentActivityName=".activities.ControlCenterv2">
```
rosenpin: Removed the parent activity because we don't want the back button to re-open the main activity, we want it to finish the current activity

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -128,6 +148,12 @@ private void validateInstallation() {
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        if (GBApplication.isDarkThemeEnabled()) {
```
rosenpin: the super.onCreate sets the theme to GadgetbridgeThemeDark with action bar, we don't want the action bar because we use a toolbar in the layout xml file

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -165,13 +191,37 @@ public void onClick(View v) {
                 setInstallEnabled(false);
                 installHandler.onStartInstall(device);
                 GBApplication.deviceService().onInstallApp(uri);
+                mProgressBar.setVisibility(View.VISIBLE);
+                IntentFilter intentFilter = new IntentFilter();
+                intentFilter.addAction("nodomain.freeyourgadget.gadgetbridge.INSTALLATION_FAILED");
+                intentFilter.addAction("nodomain.freeyourgadget.gadgetbridge.INSTALLATION_SUCCESS");
+                registerReceiver(installationResultReceiver, intentFilter);
```
rosenpin: Register the receiver to the failed/success installation intent, the receiver updates the views (progress bar/button)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -165,13 +191,37 @@ public void onClick(View v) {
...
             uri = getIntent().getParcelableExtra(Intent.EXTRA_STREAM);
         }
+
+        //Get apps count and app index from calling intent
+        int appsCount = getIntent().getIntExtra("APPS_COUNT", 1);
+        int currentAppIndex = getIntent().getIntExtra("APP_INDEX", 1);
+
+        //Set up the toolbar
+        Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
+        toolbar.setTitleTextColor(ContextCompat.getColor(getApplicationContext(), android.R.color.white));
```
rosenpin: Set the text color to white

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -165,13 +191,37 @@ public void onClick(View v) {
...
+        toolbar.setTitle(getResources().getString(R.string.installation_d_d, currentAppIndex, appsCount));
```
rosenpin: Set the title to installing ($index/$count)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -165,13 +191,37 @@ public void onClick(View v) {
...
+        toolbar.setNavigationIcon(VectorDrawableCompat.create(getResources(), R.drawable.ic_arrow_back, null));
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP)
+            toolbar.setElevation(16);
```
rosenpin: Set elevation only on lollipop+ devices (the fancy shadow)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
```diff
@@ -165,13 +191,37 @@ public void onClick(View v) {
...
+        toolbar.setNavigationOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                finish();
```
rosenpin: Finish the current installation activity when the back button is clicked

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/appmanager/AppManagerActivity.java
```diff
@@ -213,8 +215,21 @@ public void onActivityResult(int requestCode, int resultCode, Intent resultData)
...
-            startIntent.setDataAndType(resultData.getData(), null);
-            startActivity(startIntent);
+            ClipData clipData = resultData.getClipData();
+            if (clipData != null) {
+                //If there is more than one app to install
+                int appsCount = clipData.getItemCount();
+                for (int i = 0; i < appsCount; i++) {
+                    startIntent.setDataAndType(clipData.getItemAt(i).getUri(), null);
+                    startIntent.putExtra("APPS_COUNT", appsCount);
+                    startIntent.putExtra("APP_INDEX", Math.abs(i - appsCount));
```
rosenpin: Because we want it to be (1/3) -> (2/3) -> (3/3), not (3/3) -> (2/3) -> (1/3)
Why not use ```for (int i = appsCount; i >=0 ; i--)``` you ask?
That would be a good question.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleIoThread.java
```diff
@@ -625,12 +625,15 @@ void installApp(Uri uri, int appId) {
...
     private void finishInstall(boolean hadError) {
+        Intent resultIntent;
         if (!mIsInstalling) {
             return;
         }
         if (hadError) {
+            resultIntent = new Intent("nodomain.freeyourgadget.gadgetbridge.INSTALLATION_FAILED");
             GB.updateInstallNotification(getContext().getString(R.string.installation_failed_), false, 0, getContext());
         } else {
+            resultIntent = new Intent("nodomain.freeyourgadget.gadgetbridge.INSTALLATION_SUCCESS");
```
rosenpin: The installation was successful, set the intent action to success

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleIoThread.java
```diff
@@ -625,12 +625,15 @@ void installApp(Uri uri, int appId) {
     }
 
     private void finishInstall(boolean hadError) {
+        Intent resultIntent;
         if (!mIsInstalling) {
             return;
         }
         if (hadError) {
+            resultIntent = new Intent("nodomain.freeyourgadget.gadgetbridge.INSTALLATION_FAILED");
```
rosenpin: The installation was unsuccessful, set the intent action to failed

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleIoThread.java
```diff
@@ -646,6 +649,7 @@ private void finishInstall(boolean hadError) {
                 }
             }
         }
+        getContext().sendBroadcast(resultIntent);
```
rosenpin: Send the broadcast

app/src/main/res/layout/activity_appinstaller.xml
```diff
@@ -1,77 +1,69 @@
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
rosenpin: This layout is pretty much a list of views that goes from top to bottom, no reason to use RelativeLayout with so many rules when we can simply use a vertical LinearLayout

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
danielegobbetti: See above about extending AppCompat activity: I believe you will not need this after changing.
Since it took some time to get the colors under control (and is not all done yet :-) ) I really would like to avoid using colors outside of the theme-inherited ones.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
danielegobbetti: AppCompat should take care of this as well without explicit version checking.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
rosenpin: I'll change it, BTW, I'm pretty sure we should no longer use the themes that include action bars, AFAIK toolbars are the way to go.
https://medium.com/@ZakTaccardi/goodbye-actionbar-apis-hello-toolbar-af6ae7b31e5d

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
rosenpin: Got it, right

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
rosenpin: Done, I also merged the toolbars by using a toolbar.xml file, I include that file both in the ControlCenterv2 activity and the FwAppInstallerActivity activity layouts

rosenpin: I still need to check if the dark theme is enabled and change the theme if it is. But the else statement is redundant

rosenpin:
Also, if you think it's a good idea, I can easily make the selected file install automatically without the need to click the install button and then the back button every time

cpfeiffer:
Sweet! 
I just tested Mi Band firmware installation. Installation worked, but the progress notification never finishes. Might well be that this is unrelated to this PR though.

rosenpin:
Oops, that makes sense, as I only made it so the PebbleIoThread sends a broadcast when the installation finishes/fails

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/FwAppInstallerActivity.java
danielegobbetti: I agree, we didn't because of lack of time on our end, perhaps you can open a new PR after this one is merged to take care of the migration? It would be really great!

danielegobbetti:
@rosenpin I believe cf1451bde5f51678ff19b57fb9da5157a0668dcd should be reverted. As it is now the theme would remain dark until activity is destroyed / application is restarted.

rosenpin:
> I'm wondering if the multiple file selection should be limited to pbw (apps), as it would not work (but could be dangerous if it worked) for firmwares.

Good point, though it might be a bit confusing for users if the app won't explain the special behavior, maybe there should be different installations screen for apps/software?

rosenpin:
@danielegobbetti 
Unrelated question: is it currently possible to filter/blacklist certain apps from posting notifications? If it isn't I can implement this feature, if it is, please let me know how to do this because I couldn't find an option to do so

ashimokawa:
Yes there is a blacklist feature in settings

rosenpin:
Thanks!

cpfeiffer:
So how do we proceed with this?

Issue #673, opened 2017-05-01
New statistics tab


Vebryn:
#### Before opening an issue please confirm the following:
- [X ] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [X ] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.

#### Your issue is:
Some issues talk about UI refactoring.

What do you think about a new tab to show detailed speed statistics ?

<a href="https://www.hostingpics.net/viewer.php?id=233200Screenshotfrom20170501151800.png"><img src="https://img15.hostingpics.net/thumbs/mini_233200Screenshotfrom20170501151800.png" alt="Heberger image" /></a> 

On this screen, last 24h activity is sliced into 5 remarkable speeds. On April 29, I walked 54 minutes at an average speed of 21 feet per minute. I also reached a maximum speed of 105 for 2 minutes.

danielegobbetti:
Is the screenshot a mockup or have you coded the graph already? If so, I think you can open a PR and then we can discuss about details there before merging (e.g. the units to use for distance and speed :-) )

ashimokawa:
Didn't debug yet but here (s4 mini, lineage 14.1) the  tab is blank and shows no data.

Vebryn:
I notice 2 bugs:
* Y legend is displayed with 1 decimal precision
* speed >100 is wrong sorted

PR #692, opened 2017-05-15
auto-export new data after fetching (csv)

Vebryn:
Someone got no data, this commit adds debug

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
@@ -338,26 +338,34 @@ public class GBApplication extends Application {
     private static HashSet<String> blacklist = null;
 
     public static boolean isBlacklisted(String packageName) {
+        if (blacklist == null) {
+            GB.log("isBlacklisted: blacklisti is null!", GB.INFO, null);
+        }
         return blacklist != null && blacklist.contains(packageName);
     }
 
     public static void setBlackList(Set<String> packageNames) {
         if (packageNames == null) {
+            GB.log("Set null blacklist", GB.INFO, null);
             blacklist = new HashSet<>();
         } else {
             blacklist = new HashSet<>(packageNames);
         }
+        GB.log("New blacklist has " + blacklist.size() + " entries", GB.INFO, null);
         saveBlackList();
     }
 
     private static void loadBlackList() {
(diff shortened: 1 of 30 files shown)


Vebryn:
Related to #673, no #686

cpfeiffer:
I think this is not necessary anymore, since @ashimokawa fixed the issue with Pebbles.

cpfeiffer:
Besides, if you log something, it is best to use an own logger for every class (see the static LOG field in many classes). This way you get all log statements automatically prefixed with the class name for easy grepping.

Vebryn:
Understood, I'm going to review my logger. Why does travis-ci failing ? Is it my logger !?

ashimokawa:
@Vebryn 
Travis failure was random. You can check the travis log yourself, it failed to download a dependency.

That being said you should update your branch and only have one commit in the PR, not all previously merged commits again.

Also I cleaned up your code a bit (removed tons of unused stuff you only copy and pasted as was not neccessary, renamed piechart and sleep variables to something more sane etc). Please rebase your future work on that.

And please use LOG.warn() LOG.info() and LOG.error() for logging. ;)

Vebryn:
Hello,

I have switched from GB logging to Android logging. I also add a functionality to auto-export data into a csv file.

cpfeiffer:
I think this PR contains too many commits -- I think we only need 64704e872e6a532d8477deffce8adef596760d29, right?

AFAICS, this is only done for Mi Bands, atm. If we want that, it should be done for all devices. I guess we should fire an event whenever activity data is being successfully provided by a device and have somone optionally (in the service?) react on that event in order to perform the export.

Additionally, I'd like to see the export in a separate class (CSVExport?), not in the SampleProvider. Also a header comment in the CSV file would be nice for people who want to understand what it means :-)

Finally, the close() of the FileWriter should be done in a finally-block ;-)

Vebryn:
Is fetching available with other devices ? FetchActivityOperation only mentions MiBand device.

I agree with a static method in a dedicated class: CSVExport.exportToCSV. What about CSV.export ?

danielegobbetti:
Fetching is not possible for pebbles (with stock firmware and no helper app on watch).

Vebryn:
Export is not available when device doesn't support fetching. Actually, export is done after fetching. No fetching means no export.

Vebryn:
According to #709, PR closed, I'm going to open a new one into a dedicated branch.

PR #711, opened 2017-07-11
Add device s200

albertsal:
I added the s200 sport device management similar to the MAKIBESF68 ..

I modified the showIncomingCall logic, showText,
encodeStringToDevice.

The device works in unicode but for incoming calls is mixed

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusSupport.java
@@ -132,7 +132,7 @@ public class HPlusSupport extends AbstractBTLEDeviceSupport {
         gbDevice.setState(GBDevice.State.INITIALIZED);
         gbDevice.sendDeviceUpdateIntent(getContext());
 
-        if(syncHelper == null) {
+        if (syncHelper == null) {
             syncHelper = new HPlusHandlerThread(getDevice(), getContext(), this);
             syncHelper.start();
         }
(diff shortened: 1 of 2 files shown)


ashimokawa:
Thanks!

Can you please remove all the formatting changes? Whitespaces and tabs.

Also please do the same indention as we do (default Android Studio)

albertsal:
@ashimokawa sorry ... i use project settings ... can you tell me the parameters to set?

jpbarraca:
I prefer having different classes overriding the default behaviour whenever necessary. Especially in this case where the number of different devices may be large.

The rest of the code seems to be ok.

ashimokawa:
@jpbarraca 
Welcome back ;)
Do you want to merge this and then change the code according to your taste?

About the indention, it would be cool if you could use the same style as android studio default which we are using to prevent unrelated formatting changes from happening.

jpbarraca:
@albertsal How can we distinguish the S200 from the other HPLUS devices?

albertsal:
As for the MAKIBESF68 also s200 starts with SPORT

jpbarraca:
Is there an additional message supported? or a characteristic? If nothing is available we will have to resort to a configuration option.

What is the hardware and firmware versions?

albertsal:
I attach the information I get


[gadgetbridge.txt](https://github.com/Freeyourgadget/Gadgetbridge/files/1157211/gadgetbridge.txt)

jpbarraca:
I find it odd that this is required (mixed language support). Also I find not accurate method to identify the device type.
What is the problem that you find with the current version?
Also, can you please try [this version](https://github.com/Freeyourgadget/Gadgetbridge/files/1159839/app-debug.apk.zip) ?

albertsal:
For messages it requires unicode but for calls if I use unicode he writes in chinese
Ok I try your version

albertsal:
When I receive a message, the message is displayed with spaces .... I have solved this problem by applying the first 2 bytes characteristic of the unicode only at the beginning of the message the remaining characters apply only the final 0 .... do Reference to my code ...

Regarding Chinese call visualization calls (this problem is also visible with hplus) I have solved not using unicode for this function

Issue #472, opened 2016-12-24
Choose displayed items on Mi Band 2


6arms1leg:
It is possible to change the items displayed on the Mi Band 2. The standard displayed items are:
* Time/Date
* Steps
* Heart rate

But other items (e.g. battery status and calories) can be enabled, too.

Here is a BT snoop log of enabling all items to be displayed on the Mi Band 2 with the original Mi Fit app:

[btsnoop_hci-miband2_menu_all_on.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/671764/btsnoop_hci-miband2_menu_all_on.zip)

... and one disabling all again to only display the time/date:

[btsnoop_hci-miband2_menu_only_time.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/671765/btsnoop_hci-miband2_menu_only_time.zip)

If you find it worth implementing, it would be a nice feature. :bowing_man:

cpfeiffer:
Thanks! I alread have that on my TODO list, but it involves some extra UI work (like icons etc.) so I haven't done it yet.

6arms1leg:
> [...] but it involves some extra UI work (like icons etc.) [...]

Hmmm, not sure if we are talking about the same feature here. Sorry, if I wasn't specific enough. Just to be sure:
I didn't mean changing the displayed icons for the different notifications (which is also possible and a nice feature, BTW) – I meant changing the menu on the Mi Band 2. Since you mentioned UI work with icons, I wasn't sure if you meant the same. In the Mi Fit app this is implemented by selecting items in a list (with no icons), like this (for example):

[x] Time/date (always checked, since it can't be disabled)
[x] Steps
[ ] Calories
[x] Heart rate
[x] Battery

cpfeiffer:
You're right, we might get away without using icons :-)

Vebryn:
This is related to #485, both require a notification.

cpfeiffer:
Closing, thanks to @joserebelo

PR #712, opened 2017-07-12
Mi Band 2: Display item settings

joserebelo:
Allows the user to change what items are displayed on the Mi Band 2 screen.

Implements #472.


cpfeiffer:
Awesome, thanks a lot! :-)

Issue #485, opened 2017-01-01
Mi Band 2 Feature Request: Goal notification


Jojonintendo:
Hi everyone!

Here is the btsnoop of another option of the Mi Fit app, which allows the band to vibrate if the daily step goal is achieved. Thanks for your hard work!

[btsnoop_hci_goal_notification.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/679992/btsnoop_hci_goal_notification.zip)

Vebryn:
Interesting, hopelessly your btsnoop is empty.

Jojonintendo:
Hi,

If I remember correctly, I enabled and disabled the option in the Mi Fit app a few times. If the btsnoop is empty I don't really know how to provide the information needed. I would appreciate any instruction on this.

cpfeiffer:
It's not empty. You have to write 0x06 0x06 0x00 0x01 to the 00000003-... characteristic to enable the notification and 0x06 0x06 0x00 0x00 to disable it.

On success, you get 0x10 0x06 0x06 0x00 0x01 as a change notification.

I actually have all of @Jojonintendo's dumps prepared, but I really didn't find the time to add all those options. (Sorry!) So help is appreciated. I can help deciphering the dumps.

Jojonintendo:
@cpfeiffer Don't worry about all those features, they're completely optional and I love how GB works right now. I really appreciate all the effort this team makes, that's how you make such a great app!

joserebelo:
@cpfeiffer I can add this, plus the other requested options, if it is fine with you.

My only question is about how the change notification should be handled. It seems that one is being sent whenever any configuration is written to that characteristic, and they're currently not being handled for any of the options. Should we be using it to guarantee that the option was applied and and revert it otherwise? This might need an issue of its own.

cpfeiffer:
@joserebelo That's a very good question and a long standing open issue in the BLE part of Gadgetbridge. 

Since it really is a separate topic, I opened #715 and added my current thoughts about this. Discussion and suggestions much appreciated :-)

Back to this topic, I'm more than happy if you can add these options to Gadgetbridge!

cpfeiffer:
Closing, thanks to @joserebelo

PR #717, opened 2017-07-15
Mi Band 2: Goal notification

joserebelo:
Allow the user to enable or disable the daily steps goal notification (#485)


cpfeiffer:
I think we also need to set this in phase2Initialize(), but I'll merge this anyway and add the call myself.

cpfeiffer:
Thanks a lot!

Issue #483, opened 2017-01-01
Mi Band 2 Feature Request: Do Not Disturb


Jojonintendo:
Hi everyone and happy new year!

The "Do Not Disturb" mode of the Mi Fit app would be very useful, especially at night time, so that the screen won't turn on on random movements. 

It has different options: automatic (it detects if we're sleeping by itself), manual (on a set time, I believe 21:00-7:00 by default), and off. There is also the option to deactivate the screen wake by wrist movement, which is the most useful I think.

I provide a btsnoop of activating those options, and a separate one just for the screen wake. Hoping that it's useful. Thank you very much for your work!

[btsnoop_DND.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/679989/btsnoop_DND.zip)

Dako300:
I would love to see this too

mueller-ma:
I have some suggestions how to improve this feature:

1. Set indipendend times for each day of the week
2. Possibility to set "From 22:00 until I am awake"

cpfeiffer:
I think we have enough preferences already :} FWIW, newer devices like the Bip allow you to manually toggle DND mode on the device itself.

mueller-ma:
IMO the current implementation is very useless:
I dont want to be woken up by a notification, so I have to set DND from 22:00 until 11:00 to cover Mo-Fr and weekend at once

danielegobbetti:
AFAIK the DND settings of android allow to set different periods for different days of the week, and gadgetbridge can be set up to respect those settings. Or am I missing something?

mueller-ma:
I usually have my phone always muted

PR #718, opened 2017-07-15
Mi Band 2: Do Not Disturb

joserebelo:
Allows the user to change the Mi Band 2 Do Not Disturb setting (implements #483).

I've added a new TimePreference (wasn't quite sure where to put it, it's in the util package), wrapping a TimePicker into a Preference.

I've also placed the new settings in a new screen in order to avoid cluttering the main one too much (some preferences already need reordering / regrouping).

Feedback is welcome, I can fix any issues or conflicts (specially if my other PR is merged first) and rebase this PR.


cpfeiffer:
Good question where to put the TimePreference class. We can keep it in .util for now and see later if we find a better package.

Indeed, the more preferences we get, the more do we need to improve the grouping. At some point it might make sense to completely decouple the Mi1 from the Mi2 settings.

So, as you suggested, I'll have a look at merging your other PR first and see if this needs fixing afterwards.

Issue #486, opened 2017-01-01
Mi Band 2 Feature Request: Activity reminder & co


Jojonintendo:
Hi everyone!

I provide the btsnoop of yet another option in the Mi Fit app, which allows the band to tell us we have to move a little after sitting to long. One can also deactivate this option on some hours, for example at work. Thanks for your hard work!

[btsnoop_hci_activity_reminder.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/679995/btsnoop_hci_activity_reminder.zip)

Also, as to not open too many issues for not so relevant options, I provide the btsnoops for activating bluetooth visibility of the band. I don't know if it's really useful for us, but as I saw it today, I took a btsnoop too :D.

[btsnoop_hci_bt_visibility.zip](https://github.com/Freeyourgadget/Gadgetbridge/files/679996/btsnoop_hci_bt_visibility.zip)

joserebelo:
The option for activating the Bluetooth visibility is not working here (not even with the Mi Fit application), tested it on firmware 1.0.1.39 and 1.0.1.50, with the same result: the band will not be visible while connected to the phone, becoming visible when disconnected, no matter the option.

cpfeiffer:
Closing, again thanks to @joserebelo

PR #719, opened 2017-07-16
Mi Band 2: Inactivity Warnings

joserebelo:
Allow the user to enable inactivity warnings (implements #486)

The band supports 2 separate intervals for the inactivity warnings. Mi Fit has a do not disturb interval hard-coded from 12:00 to 14:00, I have made it modifiable as well.

I've also added the option to set the inactivity time threshold (which I think is the 3rd byte, it is not accurate to the minute, might need to test it a bit further, but seems to be close enough).


cpfeiffer:
Wow, you rock! :-)

PR #729, opened 2017-07-25
Greek transliteration Map and Auto connect on startup if Bt is enabled

lazarosfs:
I include 2 patches.
First is transliteration map for greek chars.
Second adds autoconnect on startup if bluetooth adapter is enabled (according to prefs)
It also tries to enable bluetooth adapter with user intervention when connect button is used, then waits for 2 secs to make sure bt is on and continues trying to autoconnect. If bt is still not enabled it break's because there is no point trying to connect with bt off.
This helps especially after rebooting the device where GB autostarts and if bt is enabled it should try to connect


lazarosfs:
Hi, thank you for the great work you 've all done.
I have made a fiew improvements and found a lot of things need tweeking mostly in Mi Band 2 support which I have.

I will try to do that but I just started with git and am a bit lost I really don't know how.
I will incorporate the new preference to commit 2. Honoring connect bt preference should also be there sure.

PR #730, opened 2017-07-26
greek transliteration map

lazarosfs:
Adds support for Greek Chars transliteration (to latin)


cpfeiffer:
Great job, thanks a lot!

PR #735, opened 2017-07-26
EnableBtOnConnect

lazarosfs:
This enables bt on connect with user intervention (adding general preference)
It also tries to connect when starting if bt is enabled


lazarosfs:
Does anyone know why checks fail?
It does compile and run fine on my pc

PR #739, opened 2017-07-28
Hungarian translation updated.

geripgeri:


cpfeiffer:
Thanks a lot!

ashimokawa:
@cpfeiffer
Please do not merge these it causes trouble. Particularly in this case I have no idea what to do when tx pulling

@geripgeri 
Please use transifex

geripgeri:
@ashimokawa Okay, sorry. I used Android Studio's built in locate editor.
Are there any further steps I should take?

ashimokawa:
@geripgeri 

No need to apologize, we are always happy about translations. Unfortunately transifex does not play well with editing files manually. Also our README says we will accept PRs (we should change that in my opinion)

Your pull request contained a lot of strings with the ids starting with  p_ but we have nothing prefixed with p_. When I pull from transifex now these will be removed (rightfully, because they are not used anyway then).

I will just pull the transifex version and maybe if you have time you can translate the missing 20 strings  here 

https://www.transifex.com/freeyourgadget/gadgetbridge/translate/#hu/strings

geripgeri:
I have finished translating the missing lines  at transifex 🙂

ashimokawa:
@geripgeri

Thanks 👍
Merged

PR #753, opened 2017-08-08
Adjust colors

mueller-ma:
I adjusted the colors a bit to
1. match material colors
2. have a visible distinction between primary_dark and primarydark_dark
3. match the color used by the action bar and side drawer in dark theme

Before
before
After
after


danielegobbetti:
Looks good to me, merging. Thanks @mueller-ma !
If you have some time and want to help regarding theming, let us know!

mueller-ma:
I dont know much about theming and gb has two good themes. The only thing is that the changelog has a white background. I will look into this

PR #762, opened 2017-08-13
Adjust strings to meet material guidlines

mueller-ma:
I adjusted the strings to meet material guidlines. Mostly it is Capitalization: https://material.io/guidelines/style/writing.html#writing-tone

Does HW stand for Hardware (revision)? I think we dont have to use a abbrivation there.

I also have replaced ... with its hexcode


ashimokawa:
Thanks!

Unfortunately Transifex would kill all translations if we do that - we need to get rid of Transifex. I would prefer if they tag translation as "possible broken" instead of removing them as soon as the original strings.xml gets modified.

We will think about how to deal with that before we merge.
Yes HW stands for Hardware revision. We used the abbreviation because the old main activity had no space for the string.

Is there a reason why you used a hex code instead of ...? I would prefer ..., also for translation.

mueller-ma:
I use the hexcode because I have seen Android Studio throwing lint errors in pther projects, however I did not check it in GB.

Weblate shows "Strings with changed source string" with a div what has changed

ashimokawa:
We should switch. Can't be worse than transifex

mueller-ma:
@ashimokawa Can you merge?

ashimokawa:
No, sorry, weblate integration does not work
The latest translations are here on github, weblate does not follow. Unfortunately.

ashimokawa:
@mueller-ma 
Seems to work now, I will merge this

ashimokawa:
@mueller-ma 
I merged this but I reverted the … change
I don't get the point of hiding …, it just puzzles me if I see a hex code.

danielegobbetti:
Can &hellip; entity be used in android XML files? Having the three dots is indeed not ideal.

mueller-ma:
They can be used, but android studio is throwing lint errors

ashimokawa:
The hex code is the worst of all three option.
Imagine the Japanese translation would consist of hex codes only.
What is the rule anyway for lint?

mueller-ma:
Cannot find it, maybe the rule has been removed

mueller-ma:
After merging this PR there is one new problem:
The strings "connecting", "connected", etc. are used at different locations, e.g. GB Push notification and on the device card in the main activity.
On is used in a sentence ("Mi Band connected") and one standalone ("connected"). This would need two different strings. I've tried to fix this in https://github.com/mueller-ma/Gadgetbridge/commit/1dbfea925890805ba487ec4e61423fcd5d676bd6 but cannot test it, because I only have a AVD for debugging.

Maybe it is better to add a paramter to https://github.com/mueller-ma/Gadgetbridge/commit/1dbfea925890805ba487ec4e61423fcd5d676bd6#diff-d5e00ac52c9ce247c3038db8572bff37L282 to get a string with first letter capitalized or not

PR #777, opened 2017-08-23
export fetched data into a csv file

Vebryn:
Hi,

Here is a clean version of csv export implementation. Old conversation can be found into #692.

Best regards

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractSampleProvider.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractSampleProvider.java
@@ -22,11 +22,19 @@ import android.support.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import de.greenrobot.dao.AbstractDao;
 import de.greenrobot.dao.Property;
 import de.greenrobot.dao.query.QueryBuilder;
 import de.greenrobot.dao.query.WhereCondition;
+
 import nodomain.freeyourgadget.gadgetbridge.database.DBHelper;
 import nodomain.freeyourgadget.gadgetbridge.entities.AbstractActivitySample;
 import nodomain.freeyourgadget.gadgetbridge.entities.DaoSession;
@@ -40,6 +48,7 @@ import nodomain.freeyourgadget.gadgetbridge.model.ActivityKind;
  * @param <T> the sample type
  */
 public abstract class AbstractSampleProvider<T extends AbstractActivitySample> implements SampleProvider<T> {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractSampleProvider.class);
     private static final WhereCondition[] NO_CONDITIONS = new WhereCondition[0];
     private final DaoSession mSession;
     private final GBDevice mDevice;
@@ -88,6 +97,29 @@ public abstract class AbstractSampleProvider<T extends AbstractActivitySample> i
     @Override
     public void addGBActivitySamples(T[] activitySamples) {
         getSampleDao().insertOrReplaceInTx(activitySamples);
+    }
+
+    @Override
+    public void exportToCSV(T[] activitySamples, File outFile) {
+        String separator = ",";
+
+        LOG.info("Exporting samples into csv file: " + outFile.getName());
+        try {
+            BufferedWriter bw = new BufferedWriter(new FileWriter(outFile));
+
+            for (T sample : activitySamples){
+            String line = sample.getTimestamp() + separator + sample.getDeviceId() + separator + sample.getUserId() + separator + sample.getRawIntensity() + separator + sample.getSteps() + separator + sample.getRawKind() + separator + sample.getHeartRate();
+
+            //LOG.debug("Adding line into buffer: " + line);
+            bw.write(line);
+            bw.newLine();
+        }
+
+            bw.flush();
+            bw.close();
+            } catch (IOException e) {
+            LOG.error(e.getMessage());
+        }
     }
 
     @Nullable
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchActivityOperation.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchActivityOperation.java
@@ -27,6 +27,7 @@ import android.widget.Toast;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.io.IOException;
 import java.text.DateFormat;
 import java.util.ArrayList;
@@ -57,6 +58,8 @@ import nodomain.freeyourgadget.gadgetbridge.service.devices.miband2.MiBand2Suppo
 import nodomain.freeyourgadget.gadgetbridge.service.devices.miband2.AbstractMiBand2Operation;
 import nodomain.freeyourgadget.gadgetbridge.util.ArrayUtils;
 import nodomain.freeyourgadget.gadgetbridge.util.DateTimeUtils;
+import nodomain.freeyourgadget.gadgetbridge.util.FileUtils;
+import nodomain.freeyourgadget.gadgetbridge.util.CSVExport;
 import nodomain.freeyourgadget.gadgetbridge.util.GB;
 
 /**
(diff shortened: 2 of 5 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractSampleProvider.java
```diff
@@ -90,6 +99,29 @@ public void addGBActivitySamples(T[] activitySamples) {
         getSampleDao().insertOrReplaceInTx(activitySamples);
     }
 
+    @Override
+    public void exportToCSV(T[] activitySamples, File outFile) {
+        String separator = ",";
+
+        LOG.info("Exporting samples into csv file: " + outFile.getName());
+        try {
+            BufferedWriter bw = new BufferedWriter(new FileWriter(outFile));
```
cpfeiffer: Please use
`try (BufferedWriter bw = ....) {`
and remove the `bw.close()`
That way the writer will always be closed (also when an exception occurs).

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractSampleProvider.java
```diff
@@ -90,6 +99,29 @@ public void addGBActivitySamples(T[] activitySamples) {
...
+
+            //LOG.debug("Adding line into buffer: " + line);
+            bw.write(line);
+            bw.newLine();
+        }
+
+            bw.flush();
+            bw.close();
+            } catch (IOException e) {
+            LOG.error(e.getMessage());
```
cpfeiffer: Please use
`LOG.error("Explain what failed", e);`
otherwise, the stacktrace of the exception will not be logged, making debugging hard.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchActivityOperation.java
```diff
@@ -207,6 +210,14 @@ private GregorianCalendar saveSamples() {
 
                     timestamp.add(Calendar.MINUTE, 1);
                 }
+
+                // export csv on fetching
+                File myPath = null;
+                myPath = FileUtils.getExternalFilesDir();
+                File myFile = new File(myPath, startTimestamp.getTimeInMillis() / 1000 + ".csv");
```
cpfeiffer: Maybe add a `gadgetbridgedb_` prefix (or something like that, maybe shorter)?

But really, this does not belong here. It should be implemented in one place for all devices, not in the Mi Band 2 FetchActivityOperation. And I guess we should also have an option to turn this on and off.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/CSVExport.java
```diff
@@ -0,0 +1,51 @@
...
+public class CSVExport {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractSampleProvider.class);
+
+    public static void exportToCSV(AbstractActivitySample[] activitySamples, File outFile) {
+        String separator = ",";
+        BufferedWriter bw = null;
+
+        LOG.info("Exporting samples into csv file: " + outFile.getName());
+        try {
+            bw = new BufferedWriter(new FileWriter(outFile));
```
cpfeiffer: See above with `try (BufferedWriter ...)` and the `bw.close()`

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/CSVExport.java
```diff
@@ -0,0 +1,51 @@
...
+            bw.write("TIMESTAMP" + separator + "DEVICE_ID" + separator + "USER_ID" + separator + "RAW_INTENSITY" + separator + "STEPS" + separator + "RAW_KIND" + separator + "HEART_RATE");
```
cpfeiffer: Shouldn't we also export the normalized values?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/CSVExport.java
```diff
@@ -0,0 +1,51 @@
...
+import nodomain.freeyourgadget.gadgetbridge.entities.AbstractActivitySample;
+
+/**
+ * Created by Vebryn on 29/05/17.
+ */
+
+public class CSVExport {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractSampleProvider.class);
+
+    public static void exportToCSV(AbstractActivitySample[] activitySamples, File outFile) {
```
cpfeiffer: I think it's sufficient to use `ActivitySample` here instead of `AbstractActivitySample`.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractSampleProvider.java
Vebryn: I didn't know this try block syntax. So, it's a try-with-resources statement. It closes bw at end of execution. I understand so why you remove bw.close().

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchActivityOperation.java
Vebryn: Right, does data from other devices differs ? Where would you place on/off feature ?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/CSVExport.java
Vebryn: Normalized values are deduced from a simple matching rule. I think correlation have to be done after exportation.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchActivityOperation.java
cpfeiffer: Data from other devices may differ, yes. So either we use normalized values (= the same data ranges for all devices) or device specific raw values, or even both.

But it would be good to have this for all devices, not just one.

Since some devices provide the data in a different way, where we have no access to all samples at once, we might call a method like `samplesAdded(device, start, end)` which would then check if CSV export is enabled, then fetch the samples from the given date range from the db (probably cached in the session already) and perform the export.

Alternatively we could integrate with the new DB export functionality and perform the CSV export repeatedly. We would just need to remember the start and end dates so that we only export the incremental changes.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/CSVExport.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+public class CSVExport {
+    private static final Logger LOG = LoggerFactory.getLogger(SampleProvider.class);
+
+    public static void exportToCSV(AbstractActivitySample[] activitySamples, File outFile) {
+        String separator = ",";
+
+        LOG.info("Exporting samples into csv file: " + outFile.getName());
+        try (BufferedWriter bw = new BufferedWriter(new FileWriter(outFile))) {
```
cpfeiffer: `FileWriter` always uses the local platform encoding, which, atm, is UTF-8 on Android. Relying on the local platform encoding will in many cases lead to a bug at some point. One has to use the extra verbose
`new OutputStreamWriter(new FileOutputStream(outFile), StandardCharsets.UTF_8))`

I've spent too much of my lifetime on encoding bugs already (I actually have a T-Shirt with the text "Schei? Encoding".

For the record, Excel' CSV import does CP-1252 (similar to ISO-8859-1), LibreOffice can handle all kinds of encodings including UTF-8.

Rahulaswani:
hI @cpfeiffer @Vebryn was this solution to easily export the activity/sleep data into csv finally implimented, if yes where can i find it ?

I went through the wikis and tried to find a way to export data from the mobile app and found the 'Database Management' option in the side menu which lets you export data but i am not sure of the format, thought it might be sql db but it isn't as DB browser for SQLite cannot recognise and open the file.

Can you help me here or point me in the right direction to be able to easily export my data in a csv file

Thanks

Vebryn:
Hi @Rahulaswani,

Export option is available on my branch [csv-export](https://github.com/Vebryn/Gadgetbridge/tree/csv-export). Feel free to contribute.

Rahulaswani:
hi @Vebryn i and @izbasit couldn't find the CSV export option in the build from you code. The export options given are exactly the same as the original project. Check screenshot from your repo below

gadgetbridge export option

Can you please help us understand from where can we trigger a csv export or if there is something else that needs to be done ?

@cpfeiffer can you help here ?

elagin:
I want to try to bring the matter to an end, considering all of the above.

PR #783, opened 2017-08-29
Initial NO.1 F1 support.

protomors:
Works: connecting, writing user data, reading firmware version and battery charge, finding device.

I already understand protocol for this device enough to implement alarms, heart rate measurement, activity and sleep data fetching. But before writing more code I would like someone to check if I am doing everything right. Never wrote software for Android before. And so I had to use code for other devices as reference without understanding it completely.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1Constants.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1Constants.java
@@ -13,5 +13,15 @@ public final class No1F1Constants {
     public static final byte CMD_DATETIME = (byte) 0xa3;
     public static final byte CMD_USER_DATA = (byte) 0xa9;
     public static final byte CMD_ALARM = (byte) 0xab;
+    public static final byte CMD_NOTIFICATION = (byte) 0xc1;
+    public static final byte CMD_ICON = (byte) 0xc3;
 
+    public static final byte NOTIFICATION_HEADER = (byte) 0x01;
+    public static final byte NOTIFICATION_CALL = (byte) 0x02;
+    public static final byte NOTIFICATION_SMS = (byte) 0x03;
+    public static final byte NOTIFICATION_STOP = (byte) 0x04; // to stop showing incoming call
+
+    public static final byte ICON_QQ = (byte) 0x01;
+    public static final byte ICON_WECHAT = (byte) 0x02;
+    public static final byte ICON_ALARM = (byte) 0x04;
 }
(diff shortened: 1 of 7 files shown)


ashimokawa:
@protomors 
Wow, thanks!
From a first quick glimpse it seems good. Will have a closer look soon.

ashimokawa:
Do you want this early work to be merged, or do you prefer improving on it first?

There is only one minor thing I noticed. You have to add your device also here:


https://github.com/Freeyourgadget/Gadgetbridge/blob/master/app/src/main/res/drawable/level_list_device.xml

protomors:
>There is only one minor thing I noticed.

Thank you. So this is why correct icon was not showing in device list.

Before merging I would like to at least add notification support (this is most needed feature for me because native app does not support them well). After that I will update this pull request.

Fetching activity data is not a problem but I will need some time to learn how to store it using greenDAO. So this probably will take more time to implement.

ashimokawa:
@protomors 
Don't worry about GreenDAO yet. We can guide you through that.

One hint:

If you know the data format of activity data then you could look at how the mi band code stores values in it's database and how the MiBandSampleProvider converts the saved raw values to normalized values.

If the format is a complicated encoded bitfield with varying meaning of bits and deciphering it is work in progress, you might want to look how we store data in the misfit database (just the raw 16 bit value) and how we decode these on the fly when reading from the database.

protomors:
Notifications are working. In this state it can be at least somewhat useful for other people with this device. I wonder how many there are among users of Gadgetbridge. Would be nice to have some feedback. And from protocol it looks like to me that NO.1 are probably using it for other devices as well (like F2). But I can't test this because I do not have them.

So, @ashimokawa if you are OK with this pull request - it can be merged. Or should I combine it into one commit and rebase first?

ashimokawa:
@protomors 
I can rebase and merge all commits or I can squash them into one commit (this works with the click of a button). It is up to you ;)

Personally I like when history is preserved, so I only squash when contributors change things back and forth and have many small or reformatting commits.

protomors:
> Personally I like when history is preserved, so I only squash when contributors change things back and forth and have many small or reformatting commits.

This is why I asked. Different authors have different preferences. I am fine with either way. Then this pull request can be merged. I will create another when finished implementing another major feature.

And about data structure. This tracker has no "activities" in the same sense as MiBand. It just returns total steps for each hour. Data structure is so simple that there is probably no reason to store raw response, as other implementations do. But will data display correctly if there will be only one record per hour and not per minute like in MiBand?

ashimokawa:
@protomors 
I *think* we assume minute data so far.
I would still just save one record per hour of course in your case, and as a quick hack we could expand that in to 60 samples with 1/60th of steps each in the F1SampleProvider - this would make it compatible with the current charts activity very quickly.

protomors:
@ashimokawa
Nice idea. First I will try to return data as is and see how charts will behave. And if this does not work I will try to interpolate.

PR #785, opened 2017-08-31
Enable approximate sleep tracking for Miband 2 with new firmwares.

Roxxor91:
This gets almost the same results as MiFit, but needs some further research.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2SampleProvider.java
```diff
@@ -47,13 +47,23 @@
...
+    // 105 - waking up but staying in bed (no steps but high raw intensity)
+    // 123 - start of sleep
+
     public static final int TYPE_UNSET = -1;
     public static final int TYPE_NO_CHANGE = 0;
     public static final int TYPE_ACTIVITY = 1;
     public static final int TYPE_NONWEAR = 3;
     public static final int TYPE_CHARGING = 6;
     public static final int TYPE_LIGHT_SLEEP = 9;
+    public static final int TYPE_LIGHT_SLEEP_NEWFW = 112;
```
mueller-ma: When was the type changed? 1.0.1.47?
Can you rename the variables to ..._1-0-1-47?

ashimokawa: No please don't. This code is shared with Amazfit Bip at least right now, and there firmware versions are different.

Roxxor91: @ashimokawa Well its another device. It may have its own file in my opinion.

ashimokawa: @Roxxor91 
Sure, if we find out that the devices are really diverging.

Up to a certain firmware both acted the same, and now it seems differen. Amazfit Bip is only one month old, so it could also be a bug that it acts up.
Since both use the same Mifit app I think they will stay very similar - but we will see.

Roxxor91: @ashimokawa Well you can join the analysis in Issue #686 and do it really detailed. For that you need to sync your amazfit with MiFit and Gadgetbridge. MiFit shows exact times when you were in which Phase, which helps a lot to try to recognize patterns.
I do not think that the Amazfit is acting up. Mifit supports the old MiBands and the MiBand 2 which already differ by now.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2SampleProvider.java
```diff
@@ -47,13 +47,23 @@
     // 12 = definitely wake up
     // 17 = definitely not sleep related
 
+    // Firmware >1.0.1.47 changed sleep detection
+    // Those activities are marked wit NEWFW
```
nickvergessen: \*with

but I agree, should use the actual number, since a new firmware can pop up and then its not `NEWFW` but `MIDDLEOLDFW` ;)

danielegobbetti:
An idea about the constants: we could add new table in the database that stores activity IDs for each (whitelisted) firmware version and use that values if they are present or fall back to the old constants for unknown firmwares.

This would allow people to experiment more easily, as for new firmwares (or for testing different values) it would be enough to modify the database instead of adding constants and recompiling the app.

Also adding support for new devices' activity decoding would be easier especially to "change a value and see what happens" scenarios.

Of course rules like "convert the 10 samples prior to this one to activity type ``XX``" are almost impossible to put in the database, *that* logic would live in the code.

I understand this is out of the scope of this PR though. :)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2SampleProvider.java
danielegobbetti: Are we sure we're not getting a "record version" in some unknown byte from the miband (and the bip)? Pebble watches send back such an information and it allows us to properly decode the data.

cpfeiffer: FWIW, I really appreciate the work done here, thanks a lot! :+1: 

To be really correct, we should check the firmware version that was used to produce the sample. Otherwise we might mix up values that have meaning X in version 1.0 and meaning Y in 1.2.

I will see how we can do this in an efficient way.

nickvergessen:
I updated my firmware last week to 1.0.53 I now exported the DB again and found the following new values which are not identified until now? apart from the ones you commented above:

* 121
* 80
* 96
* 106
* 89
* 90
* 108
* 17
* 49
* 33
* 16
* 91
* 124

nickvergessen:
Okay so I looked a bit at the raw data and summed the before and after values:
https://gist.github.com/nickvergessen/4c9f365434931e958f652693b24b02c1

There seem to be quite a few values with very little appearances in this new version...

nickvergessen:
Quick question in between, can we release a version with 112 + 122 as a quick fix? this will at least make the obvious sleeping work again?

@danielegobbetti @cpfeiffer

ashimokawa:
@nickvergessen 
I don't have a Mi Band 2 so I can't check, and personally I don't want to merge this without proper testing

kmmndr:
I've a brand new Mi Band 2 with latest firmware installed, let me know if I can help testing

ashimokawa:
I am closing this as I think I have pretty much fixed this in 9f05aff11be983a88893b652ce093f314de47257, thanks for all the input, you were really helpful

nickvergessen:
Glad to hear, lets see what happens with the next release :)

PR #792, opened 2017-09-03
NO.1 F1 steps data

protomors:
I was able to successfully fetch and save steps data from device. And they are showing correctly in the "Steps a week" tab. But graph "Your steps and activity" is empty.

Then I tried to override getAllActivitySamples to return data split by minutes instead of hours, like @ashimokawa suggested. But graph is still empty. I probably missed something important. Could you please look into my code?

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
@@ -39,7 +39,7 @@ public class No1F1SampleProvider extends AbstractSampleProvider<No1F1ActivitySam
 
     @Override
     public float normalizeIntensity(int rawIntensity) {
-        return rawIntensity / (float) 100.0;
+        return rawIntensity / (float) 8000.0;
     }
 
     @Override
@@ -68,36 +68,5 @@ public class No1F1SampleProvider extends AbstractSampleProvider<No1F1ActivitySam
     @Override
     protected Property getDeviceIdentifierSampleProperty() {
         return No1F1ActivitySampleDao.Properties.DeviceId;
-    }
-
-    @Override
-    public List<No1F1ActivitySample> getAllActivitySamples(int timestamp_from, int timestamp_to) {
-        List<No1F1ActivitySample> samples = super.getGBActivitySamples(timestamp_from - 3599, timestamp_to, ActivityKind.TYPE_ALL);
-        List<No1F1ActivitySample> outSamples = new ArrayList<>();
-
-        for (No1F1ActivitySample sample : samples) {
-            int timestamp = sample.getTimestamp();
-            int steps = sample.getSteps() / 60;
-            int leftover_steps = sample.getSteps() % 60;
-            for (int i = 0; i < 60; i++) {
-                if (i == 59)
-                    steps = steps + leftover_steps; // simplest way to not lose steps from rounding error
-                if (timestamp >= timestamp_from && timestamp <= timestamp_to) {
-                    No1F1ActivitySample newSample = new No1F1ActivitySample(
-                            timestamp,
-                            sample.getDeviceId(),
-                            sample.getUserId(),
-                            steps,
-                            sample.getRawKind()
-                    );
-                    newSample.setProvider(this);
-                    outSamples.add(newSample);
-                }
-                timestamp += 60;
-            }
-        }
-
-        detachFromSession();
-        return outSamples;
     }
 }
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
@@ -1,0 +1,68 @@
+package nodomain.freeyourgadget.gadgetbridge.devices.no1f1;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+
+import de.greenrobot.dao.AbstractDao;
+import de.greenrobot.dao.Property;
+import nodomain.freeyourgadget.gadgetbridge.devices.AbstractSampleProvider;
+import nodomain.freeyourgadget.gadgetbridge.entities.DaoSession;
+import nodomain.freeyourgadget.gadgetbridge.entities.No1F1ActivitySample;
+import nodomain.freeyourgadget.gadgetbridge.entities.No1F1ActivitySampleDao;
+import nodomain.freeyourgadget.gadgetbridge.impl.GBDevice;
+
+public class No1F1SampleProvider extends AbstractSampleProvider<No1F1ActivitySample> {
+
+    private GBDevice mDevice;
+    private DaoSession mSession;
+
+    public No1F1SampleProvider(GBDevice device, DaoSession session) {
+        super(device, session);
+
+        mSession = session;
+        mDevice = device;
+    }
+
+    @Override
+    public int normalizeType(int rawType) {
+        return rawType;
+    }
+
+    @Override
+    public int toRawActivityKind(int activityKind) {
+        return activityKind;
+    }
+
+    @Override
+    public float normalizeIntensity(int rawIntensity) {
+        return rawIntensity;
+    }
+
+    @Override
+    public No1F1ActivitySample createActivitySample() {
+        return new No1F1ActivitySample();
+    }
+
+    @Override
+    public AbstractDao<No1F1ActivitySample, ?> getSampleDao() {
+        return getSession().getNo1F1ActivitySampleDao();
+    }
+
+    @Nullable
+    @Override
+    protected Property getRawKindSampleProperty() {
+        return No1F1ActivitySampleDao.Properties.RawKind;
+    }
+
+    @NonNull
+    @Override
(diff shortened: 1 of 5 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/no1f1/No1F1Support.java
```diff
@@ -430,4 +445,48 @@ private void stopNotification() {
...
+                        samples.get(i).setRawKind(ActivityKind.TYPE_ACTIVITY);
+                        provider.addGBActivitySample(samples.get(i));
+                    }
+                    samples.clear();
+                    LOG.info("Steps data saved");
+                    if (getDevice().isBusy()) {
+                        getDevice().unsetBusyTask();
+                        getDevice().sendDeviceUpdateIntent(getContext());
+                    }
+                } catch (Exception ex) {
```
cpfeiffer: Please, don't ever swallow exceptions. At the very least, log them with 
`LOG.warn("message", ex);`
But also consider showing a toast -- I guess the user would be thankful to know that his activity data went down the hill instead of being saved in the database :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public List<No1F1ActivitySample> getAllActivitySamples(int timestamp_from, int timestamp_to) {
+        List<No1F1ActivitySample> samples = super.getGBActivitySamples(timestamp_from - 3599, timestamp_to, ActivityKind.TYPE_ALL);
+        List<No1F1ActivitySample> outSamples = new ArrayList<>();
+
+        for (No1F1ActivitySample sample : samples) {
+            int timestamp = sample.getTimestamp();
+            int steps = sample.getSteps() / 60;
```
cpfeiffer: Can you add a comment why you distribute the hourly steps to minute-based samples?

Also, I would prefer to fix the problems with hourly samples than workaround them. Can you elaborate on them?

protomors: There was a discussion about it in https://github.com/Freeyourgadget/Gadgetbridge/pull/783
This device stores total steps count for each hour. But looks like that activity chart expects data to be provided for each minute. So I tried this workaround, but it didn't worked. Either I implemented it wrong or graph not showing for other reasons. I was not able to tell for sure by looking into charts code. So fixing it is beyond my abilities.

Ideally I would like to display steps data using bar graph like in "Steps by week" tab. But this probably will require big changes to core of the program.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/no1f1/No1F1Support.java
protomors: Fixed (and for other exceptions as well).

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
cpfeiffer: We actually try hard not to rely on any sample "width" (timespan for which a sample is valid). The activity analysis does not mind and the charts actually use timestamps for each x-value, so it *should* work.

Apparently it does not, so we need to find out why. Do I understand you right that even though you provide minute-samples now, the charts don't display them?

protomors: >Do I understand you right that even though you provide minute-samples now, the charts don't display them?

You are correct. I included this commit anyway because it is possible that there are some unrelated errors which are causing this. 

Just now I tried to remove my getAllActivitySamples and see how it works again. With samples for every hour graph "Steps a week" works correctly. But graph "Speed zones" shows too high values like 2000 or 4000 steps per minute. Looks like it does not take into account distance between samples and assumes that all steps were made in one minute.

And graph "Your activity and sleep" is just empty. When I open it logcat shows

```
09-04 09:20:58.271 4497-15736/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.AbstractChartFragment: Your activity and sleep: number of samples:26
09-04 09:20:58.290 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 285
09-04 09:20:58.290 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 145
09-04 09:20:58.291 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 95
09-04 09:20:58.291 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 23
09-04 09:20:58.292 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 2066
09-04 09:20:58.292 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 4602
09-04 09:20:58.293 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 326
09-04 09:20:58.293 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 11
09-04 09:20:58.294 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 14
09-04 09:20:58.295 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 12
09-04 09:20:58.295 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.ActivityAnalysis: Adding: 22
09-04 09:20:58.298 4497-15737/nodomain.freeyourgadget.gadgetbridge I/nodomain.freeyourgadget.gadgetbridge.activities.charts.AbstractChartFragment: Your sleep: number of samples:26
09-04 09:20:58.425 4497-4517/nodomain.freeyourgadget.gadgetbridge D/OpenGLRenderer: endAllActiveAnimators on 0x7f778d7800 (RippleDrawable) with handle 0x7f778950e0
```

so it at least reading samples from the database. But nothing is displayed.

ashimokawa:
Unrelated question: Can we add your name to the README.md under "additional device support"?
If so, pick a name ;) - aliases are also welcome.

protomors:
@ashimokawa 
Of course, I will be glad. My real name is "Vitaly Svyastyn". I only recently started contributing on github so did not bother to fill my profile.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
protomors: It really was my mistake. I did not know that activity chart displays not number of steps, but activity intensity. Now i set value for rawIntensity column equal to number of steps and it shows on the chart. Can you please explain exact meaning of this value? Am i understanding right that method normalizeIntensity has to make it fit between 0 and 1?

cpfeiffer:
Very cool! :+1:  :tada:

ashimokawa:
@protomors 
I will add your name to the README.md right before the next release ;)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/no1f1/No1F1SampleProvider.java
cpfeiffer: Yeah, I'm afraid the Speed Zones page is still a little buggy.

From a quick look, I cannot see why activity and sleep would not be displayed. Needs some debugging, I guess. If you share the db, I can try to find out what's wrong. You can mail the db or a link to it to carsten@freeyourgadget.org.

And yes, the intensity is indeed a value between 0 and 1. It is merely an activity indicator. The stronger the activity, the higher the value.

protomors: @cpfeiffer 
Problem with activity not displaying was fixed in the last commit. At first I forgot to set value for intensity and this is why chart was empty. For now I set intensity value to the number of steps (because device does not provide any other data). And to normalize I simply divide it by constant value. This step can be fine tuned later.

But I still would like for samples to be displayed as bars not as narrow spikes because there is no data between them. So I send my database anyway for you to see what I mean. Maybe there is an easy way to change this.

PR #795, opened 2017-09-04
Add initial support for Teclast H30

thedroidgeek:
Scan and connection, battery level, firmware version, date and time sync
(along with some other currently hardcoded settings), notification
support, alarm support, and some more.

Oh.. and icons too!


ashimokawa:
@thedroidgeek 

Wow, thanks! Will have a closer look in the evening after work.

Two questions:

1) Can we license the icons as CC like the other current icons?

2) Can we put your name in the readme as credits for H30 support?

thedroidgeek:
1. Yep, feel free to license them as you want.

2. And sure, thanks.

ashimokawa:
I didn't see anything real bad, so I will merge this ;)
I will add your name to the README.md as right before the next release (to not confuse people, because this is not in any release yet)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/jyou/TeclastH30Support.java
```diff
@@ -0,0 +1,484 @@
...
+    @Override
+    public void onHeartRateTest() {
+        try {
+            TransactionBuilder builder = performInitialized("HeartRateTest");
+            builder.write(ctrlCharacteristic, commandWithChecksum(
+                    JYouConstants.CMD_ACTION_HEARTRATE_SWITCH, 0, 1
+            ));
+            performConnected(builder.getTransaction());
+        } catch(Exception e) {
+            LOG.warn(e.getMessage());
```
cpfeiffer: When you get an exception, please pass the exception itself to warn() or error(), like this:
`LOG.warn("some message", e);`
That way we will get the actual stacktrace in the log, not just the message.

thedroidgeek: Gotcha.

SLAP0:
Nice! Thank you very much. I almost bought a Mi Band 2 to replace my Teclast H30. The proprietary app is a nightmare. Do you have any plan what functions you'll add next?

thedroidgeek:
@SLAP0 - I sadly don't own the watch anymore, I've lost it recently while getting out from a crowded train :( - it really gets easily unhooked... But I think I've figured most of the commands out, so it shouldn't be hard for someone else to implement the missing features.

PR #796, opened 2017-09-05
Updated it (Italian) translation

LuccoJ:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)

### DIFF
--- a/app/src/main/res/values-it/arrays.xml
+++ b/app/src/main/res/values-it/arrays.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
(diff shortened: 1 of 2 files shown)


LuccoJ:
This is my first attempt to use [Stringlate](https://f-droid.org/packages/io.github.lonamiwebs.stringlate/), so if I botched it, please don't hate me and just ignore.

danielegobbetti:
Thanks, but we're having some conflicts in weblate ATM so it's possible we have to reject this. Will followup soon!

danielegobbetti:
Unfortunately this cannot be merged as is, but I'm keeping this open because I will try to extract the changed resources from your source branch. Thanks!

PR #803, opened 2017-09-06
Mi Band 2: Implement multiple button actions & handle some configuration packets

MigNov:
Hi all,
this patch enables option for multiple button actions according to the "Delay after button action". This broadcast is being sent to the intent along with button_id extra identifying how many times have the pass been done.

By pass defined number of button presses is meant, i.e.

1) if you set number of button presses to 5 and delay after match to 2000 ms (2s) and then press Mi Band 2 Button 5 times and wait 2s it will send a defined broadcast intent message with button_id 1 in the intent's Extra field.

2) if you set number of button presses to 5 and delay after match to 2000 ms (2s) and then press Mi Band 2 Button 10 times and wait 2s it will send a defined broadcast intent message with button_id 2 in the intent's Extra field as 10 is twice 5 button presses so therefore button_id is 2.

If you want to preserve old behavior and send after 5 button presses immediately you can safely set this variable (Delay after button action) to 0 which will send the broadcast intent message immediately with button_id 0.

Also there are some configuration registers fixed by handling them silently along with comments as I've been digging some documentations on the internet. Links are provided within the code comments.

Hopefully you'll find it helpful,
Michal (a.k.a. Mig)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Service.java
```diff
@@ -50,6 +50,18 @@
...
+    // Also good resource is http://muhlenxi.com/2017/05/08/iOS-Bluetooth-Low-Energy-Develop-Chapter7/ however hard readable
+    // even using Google Translate translation as I don't know Chinese
+    // Update: Translation using following URL seems to be understandable:
+    // https://translate.google.com/translate?hl=en&sl=auto&tl=en&u=http%3A%2F%2Fmuhlenxi.com%2F2017%2F05%2F08%2FiOS-Bluetooth-Low-Energy-Develop-Chapter7%2F
+    // Another good resource may be: http://allmydroids.blogspot.cz/2014/12/xiaomi-mi-band-ble-protocol-reverse.html
+    public static final UUID UUID_CHARACTERISTIC_DEV_INFO_SERIAL_NO = UUID.fromString("00002a25-0000-1000-8000-00805f9b34fb");
+    public static final UUID UUID_CHARACTERISTIC_DEV_INFO_HW_REVISION = UUID.fromString("00002a27-0000-1000-8000-00805f9b34fb");
+    public static final UUID UUID_CHARACTERISTIC_DEV_INFO_SW_REVISION = UUID.fromString("00002a28-0000-1000-8000-00805f9b34fb");
+    public static final UUID UUID_CHARACTERISTIC_DEV_INFO_SYSTEM_ID = UUID.fromString("00002a23-0000-1000-8000-00805f9b34fb");
+    public static final UUID UUID_CHARACTERISTIC_DEV_INFO_PNP_ID = UUID.fromString("00002a50-0000-1000-8000-00805f9b34fb");
```
cpfeiffer: These are standard BLE UUIDs of the Device Info Profile, and we already handle at least some of them. The PNP ID for example, is not properly handled yet. It's not really obvious how we do that, so no wonder you didn't see that.

We have a class AbstractBleProfile with several subclasses that provide a reusable implementations of several standard Bluetooth profiles/services. The DeviceSupport classes then just create the profile instances and delegate the gatt events to them.

So, if you want to improve or further implement the standard profiles, please go ahead :-)

Can you rework this PR, because the button handling is good :-)

MigNov:
Hi,
thanks, a opened a new pull request about button actions.

Mig

cpfeiffer:
Superseded by https://github.com/Freeyourgadget/Gadgetbridge/pull/811

PR #806, opened 2017-09-08
NO.1 F1: fetch sleep data.

protomors:
I am still not very confident in my code. So I will continue to make separate pull requests for each implemented feature instead of making a lot of changes at once.

Fetching and displaying sleep data now works but check to distinguish between light and deep sleep is very rudimentary. And there is no detection of awake time so total displayed sleep time is too high.

This device does not make any data analysis, all had to be done by an app. But before trying to come up with algorithm for it I want to implement fetching of heart rate data to have as much data to work with as possible.


cpfeiffer:
Looks good, great job :-)

If we can help with anything, just ask.

protomors:
@cpfeiffer Thank you. Just the same thing as with steps data. I do not like how it is displayed on the graph :) So I will open an issue and describe everything in detail there.

Issue #639, opened 2017-04-03
Controlling multiple audio playback apps


p-friedrich:
#### Before opening an issue please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.

#### Your issue is:
I use 3 different apps for audio playback: Shuttle+(1) as general music app, PocketCast(2) for podcasts and sometimes Deezer(3) for music streaming. Here is the behaviour I expected from the media control: I start (1) and start playback, it shows song information on the watch and the controls work as expected. This works. If I open (2) and start playback, (1) stops but is still in the notification tray. If I use the media controls on the watch now, I can't really predict the outcome: Sometimes (1) starts playing again, sometimes (2) just stops. The displayed media info is also not consistent and changes to the other app sometimes. If playback is stopped and I press play on the watch, I can't be sure what starts playing. Interestingly, sometimes when playback is stopped it shows the info of (1) but pressing play starts (2). The best case scenario would be, if it remembered the last app playing and the media control always act on this app. So pressing stop always stops the playback and doesn't start another app.
I hope I was able to express my problem. If I can help to resolve this, please feel free to tell me.

#### Your wearable device is:

Pebble Time v4.3

#### Your android version is:

7.1.1

#### Your Gadgetbridge version is:

0.18.4

*New issues about already solved/documented topics could be closed without further comments. Same for too generic or incomplete reports.*

danielegobbetti:
Can you please share which preferred music app is set in gadgetbridge? Does the behavior change if you switch the configured app?

danielegobbetti:
(Marking this as device:pebble for clarity even though theoretically any gadget would behave the same)

p-friedrich:
Currently the preferred music app is set to "Standard". I will try to test different settings soon.

danielegobbetti:
Closing for the lack of feedback, please feel free to reopen once you performed your tests.

Further, if you can build from source, you could give the notification-refactor branch a try as something was changed in the logic.

bluuub:
I experience the same issue:
Using Material Player (Audiobooks) and Vanilla Music (Music). 

Test 1:
Default: Material Player

Playing: Vanilla Music

On Pebble: Pause
[...]

PR #807, opened 2017-09-08
Support control of multiple media playback applications

gabethecabbage:
Using the MediaSessionManager we can detect current playback sessions and getPackageName allowing playback control intents to be directed at the application currently playing media/last seen playing media.

This API feature is LOLLIPOP or greater only, however KITKAT functionality has been left unchanged.
On 5+ the ```Preferred Audioplayer``` preference will only be used if no playback sessions are detected.

Fixes [639](https://github.com/Freeyourgadget/Gadgetbridge/issues/639)

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/receivers/GBMusicControlReceiver.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/receivers/GBMusicControlReceiver.java
@@ -18,17 +18,24 @@
 package nodomain.freeyourgadget.gadgetbridge.service.receivers;
 
 import android.content.BroadcastReceiver;
+import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.media.AudioManager;
+import android.media.session.MediaController;
+import android.media.session.MediaSessionManager;
 import android.os.SystemClock;
+import android.util.Log;
 import android.view.KeyEvent;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.List;
+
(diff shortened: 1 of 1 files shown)


gabethecabbage:
As an additional note, this replicates pretty accurately, the behavior in the official Pebble client.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/receivers/GBMusicControlReceiver.java
```diff
@@ -73,6 +80,23 @@ public void onReceive(Context context, Intent intent) {
...
+                    audioPlayer = controller.getPackageName();
+                } catch (IndexOutOfBoundsException e) {
+                    System.err.println("IndexOutOfBoundsException: " + e.getMessage());
+                }
+            }
+
+            Log.d("keypress", musicCmd.toString());
+            Log.d("sent to", audioPlayer);
+
             long eventtime = SystemClock.uptimeMillis();
```
cpfeiffer: I didn't try it, but it looks mostly good. Two things:
- could you please make all this a separate method `private String getAudioPlayer()`
- don't use System.*.println() and don't use android.util.Log. Instead use the slf4j LOG instance, and pass a message (and the exception, if any)

Thanks! :-)

danielegobbetti:
I tested the branch and it looks good. I pushed a further commit to address @cpfeiffer remarks, will merge as soon as travis is green.

cpfeiffer:
Sorry, but this does not address my comments!?

danielegobbetti:
Sorry I misread. I'll perform the missing further changes.

PR #810, opened 2017-09-10
Add support for EXRIZU K8 smartband.

Quallenauge:
The device is compatible to HPLUS protocol.
Working:
- Alarm
- Fetch data from the device, such as heart rate, stored measurement data, stepcount, etc.
Not working:
- lost device command
- blood pressure (how to integrate these values?)

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/DeviceType.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/DeviceType.java
@@ -37,6 +37,7 @@ public enum DeviceType {
     LIVEVIEW(30, R.drawable.ic_device_default, R.drawable.ic_device_default_disabled),
     HPLUS(40, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
     MAKIBESF68(41, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
+    EXRIZUK8(42, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
     NO1F1(50, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
     TECLASTH30(60, R.drawable.ic_device_h30_h10, R.drawable.ic_device_h30_h10_disabled),
     TEST(1000, R.drawable.ic_device_default, R.drawable.ic_device_default_disabled);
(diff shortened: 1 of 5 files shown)


danielegobbetti:
Question, does the device really have  a blood pressure monitor or is it "deriving" blood pressure with the (usual) HR led monitor?

Quallenauge:
It has the HR led monitor, so the measurement should be weak in my opinion. But it is reporting them - so it would be nice if these values are shown on device screen (beside of the heart rate).

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/DeviceType.java
```diff
@@ -37,6 +37,7 @@
     LIVEVIEW(30, R.drawable.ic_device_default, R.drawable.ic_device_default_disabled),
     HPLUS(40, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
     MAKIBESF68(41, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
+    EXRIZUK8(40, R.drawable.ic_device_hplus, R.drawable.ic_device_hplus_disabled),
```
cpfeiffer: I think you mean 42 here. The keys must be unique, otherwise DeviceType.fromKey() would not work.

cpfeiffer:
I forgot to add: great job! (Also of @jpbarraca) :+1:

Quallenauge:
Thanks for your comments.
I reworked the pull request and applied your suggestion.
I also changed the setSIT() call. It doesn't harm the device.

As for the blood pressure, the values can be retrieved by getting the following data:
```diff
diff --git a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordRealtime.java b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordRealtime.java
index 4f8e4f9b..a3059704 100644
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordRealtime.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/hplus/HPlusDataRecordRealtime.java
@@ -100,6 +100,8 @@ class HPlusDataRecordRealtime extends HPlusDataRecord {
                 activityKind = HPlusDataRecord.TYPE_REALTIME;
             }
         }
+        int bloodMeasSbp = (data[10] & 0xFF);
+        int bloodMeasDbp = (data[12] & 0xFF);
     }
 
     public void computeActivity(HPlusDataRecordRealtime prev){

cpfeiffer:
Thanks! :+1: Are you willing to do some blood pressure comparisons with "real" device? It would be very interesting to know how close the values are.

Quallenauge:
If I have the opportunity to compare with a "real" blood pressure measurement device, I will post the results.
BTW: Thanks for merging =)

PR #811, opened 2017-09-10
Mi Band 2: Implement multiple button actions

MigNov:
Hi all,
this patch enables option for multiple button actions according to the "Delay after button action". This broadcast is being sent to the intent along with button_id extra identifying how many times have the pass been done.

By pass defined number of button presses is meant, i.e.

if you set number of button presses to 5 and delay after match to 2000 ms (2s) and then press Mi Band 2 Button 5 times and wait 2s it will send a defined broadcast intent message with button_id 1 in the intent's Extra field.

if you set number of button presses to 5 and delay after match to 2000 ms (2s) and then press Mi Band 2 Button 10 times and wait 2s it will send a defined broadcast intent message with button_id 2 in the intent's Extra field as 10 is twice 5 button presses so therefore button_id is 2.

If you want to preserve old behavior and send after 5 button presses immediately you can safely set this variable (Delay after button action) to 0 which will send the broadcast intent message immediately with button_id 0.

Hopefully you'll find it helpful,
Michal (a.k.a. Mig)


MigNov:
This is a rework of my original patch to cover only button actions.

Mig

cpfeiffer:
Thanks a lot, great job! :+1:  :smiley:

MigNov:
Hope you'll and also others find it useful :-)
Mig

Issue #815, opened 2017-09-14
Allow Automatic Notification Colors on Pebble


KazWolfe:
#### Before opening an issue please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:

Related to #120. The Pebble native firmware/app will tailor the background color of notifications by the app color, failing back to red if one can't be found. See an example from Discord:

![Pebble Demo](https://user-images.githubusercontent.com/5192145/30411700-72da2256-98c7-11e7-871f-eb4ebbe0aa1d.png)

Note that the Pebble Time (running from the native Android Pebble app) has a light blue app from Discord, while the Gadgetbridge version has the generic red color.

It appears though that the Pebble client has method to get the app icon for each app posting notifications. From this, it pulls the dominant icon color and then uses that in the background.

It would be nice if Pebble devices through Gadgetbridge integrated the same behavior - namely, allowing all apps to use custom background colors unless a manual color already exists for the app. If the icon color can't be retrieved, the system can use the generic red.

#### Your wearable device is:

Pebble Time (`snowy`), version 4.3. Applies to all Pebble Times

*Please specify model and firmware version if possible*

#### Your android version is:

Android 7.0.0

#### Your Gadgetbridge version is:

v0.20.0

*New issues about already solved/documented topics could be closed without further comments. Same for too generic or incomplete reports.*

ashimokawa:
@KazWolfe 
Thanks! I was not aware of this feature since I never used the official app.

I also edited your text and removed the reference to the "class" file. If we implement that we have to be legally safe and do that on our own and not look at the pebble (compliled) class file.

KazWolfe:
Noted, my bad.

Either way, still a cool feature to implement.

danielegobbetti:
[...]

PR #816, opened 2017-09-14
Update Pebble Notification Icons

KazWolfe:
**Change Contents:**

- Add a large number of notification icons missing from current version
- Update colors for existing icons to match Pebble colors exactly
- Add hooks for new icons
- Added new "Generic Phone" notification type

**Linked issue:** N/A

**Reason For Changes:**
Changes were made to add additional common icons and to add all icons from Pebble app into Gadgetbridge. Colors were also updated to match Pebble's coloring scheme exactly.

Issue #695, opened 2017-05-18
Mark notification messages as read


foosinn:
Dismissed notifications on the Pebble seems to not mark them as read in the conversation. The stock Pebble app seems to be able to do this. Also the long-press middle button to dismiss does not work.

I tested this with Whatsapp and K9.

Your wearable device is: Pebble Time Steel
Your android version is: 7.1.2 Linageos without GAPPS
Your Gadgetbridge version is: 0.19.2

IzzySoft:
Must be specific to your setup – works fine here with Pebble Time Steel, GB and K9 (cannot speak for WA, as I don't use that), just tried it out to verify. Difference here: not Lineage, but I'm still on CyanogenMod. No GApps either (I'm using microG).

foosinn:
did a quick retest. Still the notifications go away, but the messages are not marked as read. 

I'm running firmware 4.3 on the pebble. I only have "enable untested features" enabled, nothing else. 

I do not have microG. May that required? 

Am 18. Mai 2017 15:15:28 MESZ schrieb Izzy <notifications@github.com>:
>Must be specific to your setup – works fine here with Pebble Time
>Steel, GB and K9 (cannot speak for WA, as I don't use that), just tried
>it out to verify. Difference here: not Lineage, but I'm still on
>CyanogenMod. No GApps either (I'm using microG).
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/Freeyourgadget/Gadgetbridge/issues/695#issuecomment-302400097

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

gappleto97:
For me they do not even get dismissed from the notifications menu on the phone

IzzySoft:
@foosinn Sorry, I obviously misunderstood you. Messages are not marked read here either, it's just the notification on the Android device that gets dismissed along. Not sure if "mark read" is implemented at all with GB.

foosinn:
@IzzySoft may it sould be implemented as a new action "mark read"?

It would also be great if wie could assing either "mark read" or "dismiss" to long-press middle button.

IzzySoft:
@foosinn both nice ideas (though to be useful to me personally, I must be able to define that per app). I'm only an enthusiast user of GB, the decision on this must be left to the team :smile_cat:

danielegobbetti:
[...]

PR #819, opened 2017-09-15
Add support for dynamic Pebble background colors

KazWolfe:
## Changes
- Add a couple additional icon types
  - Add Lighthouse (currently unused)
  - Add Transit (public transportation app)
- Tweak the colors on existing icon types
- Implement logic to grab primary (vibrant) color from app logo
  - The color will be used when displaying a notification for an app that does not have any configs bound to it.
- Alter NotificationType to support a color (named pebbleColor)
- Alter the Pebble notification poster to listen to the color from the notification
- Alter the DeviceCommunicationService to allow for color passthrough.
- Add logic to convert HEX or Integer representations of RGB888 colors to Pebble RGB222 format.

## Linked Issues
* #815: **Allow Automatic Notification Colors on Pebble**

## Reasoning For Commit
See issue above. This commit will add in support for dynamic background colors for Pebble notifications. It has been tested on live hardware with a Pebble Time and a Pebble Time Round, and certain aspects are covered by automated tests to ensure code quality.

There should be no (known) regressions to the system due to this code.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/NotificationType.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/NotificationType.java
@@ -35,7 +35,7 @@ public enum NotificationType {
     GENERIC_EMAIL(PebbleIconID.GENERIC_EMAIL, PebbleColor.Orange),
     GENERIC_NAVIGATION(PebbleIconID.LOCATION, PebbleColor.Orange),
     GENERIC_PHONE(PebbleIconID.DURING_PHONE_CALL, PebbleColor.JaegerGreen),
-    GENERIC_SMS(PebbleIconID.GENERIC_SMS, PebbleColor.IslamicGreen),
+    GENERIC_SMS(PebbleIconID.GENERIC_SMS, PebbleColor.VividViolet),
     GMAIL(PebbleIconID.NOTIFICATION_GMAIL, PebbleColor.Red),
     GOOGLE_HANGOUTS(PebbleIconID.NOTIFICATION_GOOGLE_HANGOUTS, PebbleColor.JaegerGreen),
     GOOGLE_INBOX(PebbleIconID.NOTIFICATION_GOOGLE_INBOX, PebbleColor.BlueMoon),
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/NotificationType.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/NotificationType.java
@@ -23,7 +23,7 @@ import nodomain.freeyourgadget.gadgetbridge.devices.pebble.PebbleIconID;
 public enum NotificationType {
 
     // TODO: this this pebbleism needs to be moved somewhere else
-    UNKNOWN(PebbleIconID.NOTIFICATION_GENERIC, PebbleColor.Red),
+    UNKNOWN(PebbleIconID.NOTIFICATION_GENERIC, PebbleColor.DarkCandyAppleRed),
 
     AMAZON(PebbleIconID.NOTIFICATION_AMAZON, PebbleColor.ChromeYellow),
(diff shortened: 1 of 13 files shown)


danielegobbetti:
Great! Thanks, I will review ASAP.

KazWolfe:
Added commit 0033cef to this PR.

## Changes

- Very simple patch to make the package name retrieved lowercase.

## Linked Issues
None

## Reason
Android packages tend to be lowercase, and we shouldn't care about case when matching. It doesn't seem too important, and I'm not aware of any apps where a case difference *actually* matters.

One such affected app, however, is Slack. The code is looking for packages named `com.slack` for Slack notifications, but the actual name of the app is `com.Slack`.

This case-insensitive behavior *appears* to be used by the official Pebble app, though really it's up to GB if we want it too. However, I don't see a reason not to.

danielegobbetti:
BTW, perhaps you can take a look at #695 ? 😉

I mean the parts about handling the long push of the middle button on the pebble watch.

aveao:
The issue wasn't linked but this is about #815, btw.

danielegobbetti:
Tested and confirmed working, thanks a lot!

Before we merge, please run "Code" -> "Optimize imports" on your changed files and push the changes.

P.S. In order to test I had to comment the functionality that uses the hardcoded colors, as I don't have unsupported apps on my device :)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/NotificationType.java
```diff
@@ -35,7 +35,7 @@
     GENERIC_EMAIL(PebbleIconID.GENERIC_EMAIL, PebbleColor.Orange),
     GENERIC_NAVIGATION(PebbleIconID.LOCATION, PebbleColor.Orange),
     GENERIC_PHONE(PebbleIconID.DURING_PHONE_CALL, PebbleColor.JaegerGreen),
-    GENERIC_SMS(PebbleIconID.GENERIC_SMS, PebbleColor.VividViolet),
+    GENERIC_SMS(PebbleIconID.GENERIC_SMS, PebbleColor.IslamicGreen),
```
danielegobbetti: Why the change for SMS?

aveao: Official App uses `IslamicGreen` for messages, not `VividViolet`.

KazWolfe: Effectively what @aveao said.

I noted through testing that a value was unexpectedly violet, so I changed the SMS field. This is a follow-up of sorts to #816.

KazWolfe:
@danielegobbetti I have a commit here standing by with IntelliJ's idea of import optimization. This is global, and is changing 226 files (none of which are already in this change).

If I remember correctly, my own IDE is set to auto-optimize imports of altered files.

If you still want me to push, let me know.

danielegobbetti:
Thanks! I still don't get why the imports in DeviceCommunicationService are reordered like they are... Perhaps it's a different convention in android studio / intellij idea. 
If you want to try another round of optimization on that file, I'd appreciate that, otherwise I will merge manually tomorrow (I just chose code -> optimize imports earlier today when testing your code and all the imports were correct again, so I can do it really quickly)

KazWolfe:
Re-running optimize imports on that file returns the same result.

I'm just using the generic [`AndroidStyle.xml`](https://github.com/android/platform_development/blob/master/ide/intellij/codestyles/AndroidStyle.xml), and this is what I'm getting.

If you want, I can just manually move the `import static`s down below the dynamic imports, as that's really the only big difference I can see.

danielegobbetti:
I let optimize imports run and reverted the change of the SMS color (better preserve the continuity for Gadgetbridge users instead of official app newcomers).

KazWolfe:
Would it be possible for me to get a screenshot or similar of your import config so I can properly configure my IDE? I'd rather not go through these hoops of trying (and failing to) optimize imports again.

PR #823, opened 2017-09-21
Mi Band 2: Find lost device while in DND

### LINKED ISSUE #752: Mi2: Find lost band not working, when DND is enabled and active

#### Your issue is:
When I enable dnd times on my mi band 2 and tap on the "find band" icon during these dnd times, my band doesnt vibrate

#### Your wearable device is:
Mi Band 2
FW: 1.0.1.47
HW: v0.1.3.2

#### Your android version is:
Android 7.1.2

0.19.4

joserebelo:
This fixes #752 

While in Do Not Disturb, the Mi Band 2 doesn't vibrate for notifications / calls. If we set the alert level to vibrate only it will vibrate while in DND.

If the Amazfit Bip has the same problem we might be able to fix it in the same way.


joserebelo:
Actually, a better approach might be creating a new `AlertCategory.CustomVibrateOnly` so that a SimpleNotification could still be used, and the band would vibrate repeatedly.

Disregard this PR, at least for now.

PR #824, opened 2017-09-21
Mi Band 2: Vibrate-only notifications, find lost device while in DND

### LINKED ISSUE #752: Mi2: Find lost band not working, when DND is enabled and active

joserebelo:
Fixes #752

While in Do Not Disturb, the Mi Band 2 doesn't vibrate for notifications / calls. If we set the alert level to vibrate only it will vibrate while in DND.

Created a new custom AlertCategory, CustomVibrateOnly.

If the Amazfit Bip has the same problem we might be able to fix it in the same way, if it supports this alert level.


cpfeiffer:
Can we detect if the band is in DND mode, e.g. by reading the characteristic value? Because the display lighting up sometimes helps finding the band, e.g. in a noisy environment or for deaf people.

N.B: The ble.profiles package (here the AlertCategory enum) was supposed to implement the BLE standard only, so that we can extract the implementation to a separate repository, at some point. I guess we should just not use an enum, but plain int values for extensibility.

joserebelo:
Don't know if we can (I'm actually still not quite as comfortable with bluetooth characteristics as I'd like).

However, as a workaround we could check the preference to see if DND is enabled, and if so send the proper notification according to if we are currently in the DND time interval or not. The only problem will be when the DND is in auto mode and currently active without our knowledge (during which time the band shouldn't need to be found? Since auto mode is only supposed to enable DND while the user is sleeping, i.e. wearing it).

And yes, I didn't want to add another custom AlertCategory value, but it seemed to be the easiest way of doing it at the time.

PR #825, opened 2017-09-22
Mi Band 2: Change distance unit

joserebelo:
Allow the user to change the distance unit displayed on the band.

Found it in Mi Fit settings, hadn't noticed it before. It's a simple setting / implementation.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Coordinator.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Coordinator.java
@@ -37,6 +37,7 @@ import java.util.Set;
 
 import nodomain.freeyourgadget.gadgetbridge.GBApplication;
 import nodomain.freeyourgadget.gadgetbridge.R;
+import nodomain.freeyourgadget.gadgetbridge.activities.SettingsActivity;
 import nodomain.freeyourgadget.gadgetbridge.devices.InstallHandler;
 import nodomain.freeyourgadget.gadgetbridge.devices.SampleProvider;
 import nodomain.freeyourgadget.gadgetbridge.devices.miband2.MiBand2FWInstallHandler;
@@ -187,6 +188,16 @@ public class MiBand2Coordinator extends MiBandCoordinator {
         }
 
         return new Date();
+    }
+
+    public static MiBandConst.DistanceUnit getDistanceUnit() {
+        Prefs prefs = GBApplication.getPrefs();
+        String unit = prefs.getString(SettingsActivity.PREF_MEASUREMENT_SYSTEM, GBApplication.getContext().getString(R.string.p_unit_metric));
+        if (unit.equals(GBApplication.getContext().getString(R.string.p_unit_metric))) {
+            return MiBandConst.DistanceUnit.METRIC;
+        } else {
+            return MiBandConst.DistanceUnit.IMPERIAL;
+        }
     }
 
     public static DoNotDisturb getDoNotDisturb(Context context) {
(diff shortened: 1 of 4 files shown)


joserebelo:
@

joserebelo:
I'm assuming this also works on the Bip. @ashimokawa could you confirm?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Coordinator.java
```diff
@@ -47,6 +47,8 @@
 import nodomain.freeyourgadget.gadgetbridge.model.DeviceType;
 import nodomain.freeyourgadget.gadgetbridge.util.Prefs;
 
+import static nodomain.freeyourgadget.gadgetbridge.GBApplication.getContext;
```
ashimokawa: Thats not good

ashimokawa:
@joserebelo 
It did not work on the Bip when I last tried. That was why I did not commit similar code I wrote.

Anyway, I would prefer to consolidate the metric/imperial setting for all devices into one global setting. I know we have one for HPlus already and the pebble can also do that in theory.

I don't think it is important to have in configurable per-device(?)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Coordinator.java
joserebelo: Was just following the HPlusCoordinator, from where I based my code. Probably should have stuck to the MiBand2Coordinator style though.

joserebelo:
@ashimokawa Yes, that makes sense, I see no reason to have it configurable per-device as well.

This PR should probably be closed then, and done properly as a global setting.

protomors:
>Anyway, I would prefer to consolidate the metric/imperial setting for all devices into one global setting. I know we have one for HPlus already and the pebble can also do that in theory.

@ashimokawa Device I am working on also has the setting for displaying imperial/metric units. I just started implementing custom settings page for it. But if this setting will be made global (I too think that this is better way) - then I will wait for it.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Coordinator.java
ashimokawa: please just call GBApplication.getContext() and remove the import :)

ashimokawa:
@joserebelo 
Moved to global settings, will you adapt the PR? 
Also the name in the commit is strange, it says something about DND instead of switching measurement system..
Meanwhile switching to miles also works on Bip with Firmware 0.0.9.14, I just tested it :+1: 

@protomors 
Done, PR welcome ;)

joserebelo:
@ashimokawa I think I messed up when I rebased it (might be mixed with my other PR), I will fix them tomorrow.

ashimokawa:
@joserebelo 
@protomors 
Just pushed 6d8ffad55cecc1da3fd6bc9a04d60605070260cd , some stuff can be removed now from your PR (the preferece related stuff), and in onSendConfiguation() you have to replace the constant and use SettingsActivity.MEASUREMENT_SYSTEM

ashimokawa:
Are you still planning to update the PR, or shall I do it?

joserebelo:
I have been a bit busy, but I will finish this today.

joserebelo:
@ashimokawa I have rebased and fixed the PR.

By the way, it seems there are some localization problems. The preference summary will be correctly translated when the SettingsActivity loads, but not when changed (it changes to the preference value itself, instead of the localized string).

ashimokawa:
@joserebelo 
Thanks!

And thanks for the hint regarding the translation - I will look into into it.

PR #829, opened 2017-09-30
NO.1 F1: Display settings.

protomors:
Measurement system for distance uses new global setting. And time format is set according to Android settings (the same implementation as for MiBand).

But I have a question. I have to reconnect to band for settings to take effect because they are sent during initialization. Is this just Gadgetbridge limitation? Or do I need to implement something for settings to be sent immediatly?

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/no1f1/No1F1Support.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/no1f1/No1F1Support.java
@@ -35,6 +35,7 @@ import java.util.UUID;
 
 import nodomain.freeyourgadget.gadgetbridge.GBApplication;
 import nodomain.freeyourgadget.gadgetbridge.R;
+import nodomain.freeyourgadget.gadgetbridge.activities.SettingsActivity;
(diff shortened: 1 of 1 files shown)


ashimokawa:
You can have a look at

https://github.com/Freeyourgadget/Gadgetbridge/blob/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBandPreferencesActivity.java

There configuration changes will be sent to the device immediately.
We need something like this for the general configuration activity (I can prepare this today)

You would then have to implement onSendConfiguration() in your devicesupport class (look at MiBand2Support.java)

protomors:
Thank you. I looked into code for MiBand but didn't notice this method. Soon I will add it to this pull request.

ashimokawa:
@protomors 
Just pushed 6d8ffad55cecc1da3fd6bc9a04d60605070260cd , you can now implement onSendConfiguration in a similar fashion

ashimokawa:
I would like to release today at 23:00 CEST since the Mi2/Bip sleep fixes are important, shall I merge this as is before if there are no changes to this PR?

protomors:
I made changes to the pull request. Now it can be merged.

ashimokawa:
Thanks ;) - Will merge when i am at home

PR #838, opened 2017-10-08
NO.1 F1: Alarms support.

protomors:


cpfeiffer:
Cool, great job :-)

Do you know how many alarms the F1 supports? We want to lift the 3-alarms-limit soon.

protomors:
@cpfeiffer It has only 3. Out of curiosity I tried setting alarms with bigger indexes but it had no effect.

PR #881, opened 2017-11-11
Mi Band 2: whitelist firmware 1.0.1.54

DavisNT:
Firmware version 1.0.1.54 is marked as recommended on wiki (Mi Band
Firmware Information).
And it also seems to be working fine on my device. :)


ashimokawa:
I did not mark that as recommended. AFAIK it has problems with deep sleep. It is the firmware's fault, just we do not know yet how to postprocess the data properly to detect deep sleep. See #686

PR #885, opened 2017-11-14
Combine multipart SMS message into messages by sender

azuwis:
When receive SMS message, instead of generating multiple notifications by
PDU size, after this change, there will be only one notification for
each sender.


ashimokawa:
Thanks! Didn't have a closer look yet, but will do soon!

azuwis:
Anything needs change or explanation?

cpfeiffer:
No, looks good! A tiny difference is that multiple notifications from different senders may not preserve their order, due to the use of HashMap. A LinkedHashMap would keep the order.

azuwis:
Updated with LinkedHashMap.

ashimokawa:
I didn't know it was possible that multiple pdus from  from different senders can come it at the same time. Will merge anyway. Thanks!!

PR #894, opened 2017-11-22
[RFC] Mi Band 2: Add rudimentary raw sensor data support

atalax:
I have been playing with a third party application whose name seems better not to mention. After replaying some of the writes it does when cooperating with Sleep as Android I got something which looks very much like raw sensor data.

A logcat dump is available on https://p.atx.name/uhuhozuvoy . The bytes look easy to convert at first glance (maybe even the same as the Mi Band 1?), but I have ran out of time for today.

This is not really useful at the moment, but eventually it would be nice to add [integration](https://sleep.urbandroid.org/documentation/developer-api/) with Sleep as Android (or an OSS alternative, if there is any) (thus bypassing the need for having Mi Fit + the aforementioned third party app installed).

I am mostly putting this PR here as a request for comments/patches, so if anyone has more empty timeslots than me...

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Service.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBand2Service.java
@@ -35,8 +35,8 @@ public class MiBand2Service {
     public static final UUID UUID_CHARACTERISTIC_FIRMWARE_DATA = UUID.fromString("00001532-0000-3512-2118-0009af100700");
 
     public static final UUID UUID_UNKNOWN_CHARACTERISTIC0 = UUID.fromString("00000000-0000-3512-2118-0009af100700");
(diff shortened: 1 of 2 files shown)


custanator:
Hi! 
{0x01, 0x01, 0x19} - its accelerometer raw data
{0x01, 0x02, 0x19} - its heart rate raw data. after you should run continuous measure heart rate

thePanz:
@atx any news about this feature? you mention that "the bytes look easy t convert", do you have more details about it?

arielroque:
Has anyone that got raw sensor data from miband2 ? if yes, could show me how got it?

fredyshox:
I’ve parsed data using mentioned method from Mi Band 1 and have tested measurements of Z axis for band laying on a table. As far for positive measurements, values as expected differ by +/- 10% from value of G. But strange thing happens for negative measurements, because error rate increases to 20-30%. As I observed this happen to negative values for every axis.

Scale factor used: 250
Example values for Z axis:
recto: 10.88 m.s^-2 (~10% from G)
verso: -6.6 m.s^-2 (~30% from G)

Has anyone met his issue?

PR #895, opened 2017-11-27
Added Etar as calendar app type

tradej:
Etar is a FOSS calendar app distributed on [F-Droid](https://f-droid.org/packages/ws.xsoh.etar/). I have not tested this change since I don't have the SDK set up.


cpfeiffer:
Thanks! I haven't tested it either, but it's not going to break anything, so please test with the next version :-)
Happy New Year!

PR #944, opened 2018-01-18
recognize telegram beta notifications

NotAFile:
There are at least two more telegram apps that could be added, like plus messenger, but I can't find the application ID of those right now.


cpfeiffer:
Thanks :-)

NotAFile:
How often do you usually release App updates? I.e. is it worth it for me to set up and build this app to have the changes sooner?

Avamander:
@NotAFile With FDroid, I've understood it's pretty much up to FDroid to release a new update (even the latest update haven't been released yer), so it's quite worth it.

cpfeiffer:
We used to release about once a week, but have slowed down a bit. Right now it's indeed fdroid that failed to build our latest release.

cpfeiffer:
If you self-compile, have a look at #945 though.

PR #946, opened 2018-01-19
Recognize some more alternative apps

NotAFile:
* protonmail email
 * challegram and plus messenger Telegram clients


cpfeiffer:
Thanks again :+1:  :-)

Issue #480, opened 2016-12-31
Presence in Google Play Store


henningvs:
After doing some research I could not find any information about why the app has not been published to the "official" Android Playstore.

Is there a reason why this hasn't been done and if yes/no could I help to resolve it?

danielegobbetti:
The lack of feedback is due to the fact that we are discussing this internally.

My **personal** take on the matter is:
- on the one hand, I like the idea of promoting f-droid or anyway an un-googled android experience by providing some features outside of the "walled garden"
- on the other hand, I am the first person that discourages non tech-savy from allowing untrusted sources on their devices

IzzySoft:
I partly agree with @danielegobbetti here:

* full ack to his first bullet-point. Adding to that: with GB not available there, pebblers now should be more tempted to give F-Droid a try :smile_cat: 
* partly ack to the second one. AFAICS, that "unknown sources" main intention is to keep people glued into the Google ecosystem. Even with it enabled, the user has to confirm side-loaded installations via the package manager (I'm not speaking of `adb install` here, which disregards that setting anyhow). To my knowledge, only system apps can bypass that verification, as others don't get the `INSTALL_PACKAGES` permission granted. And if some malware managed to break into that, our nice "protection setting" has long become irrelevant anyhow.

My additional, personal bullet-points would be:

* the excuse of "I cannot access that store" doesn't pull here. There are regions (and other reasons) where/why people cannot access Playstore, but F-Droid is open to everyone.
* don't "feed the trolls" (no offense meant, of course). I know it's quite cosy with all that Google stuff, Facebook, WhatsApp & Co. But we're in the "Privacy first" region here. Forcing people to open their eyes might help them to, well, open their eyes as they have no other choice – if you know what I mean.

So I, personally, see no reason why GB should be available at Google Play. But as emphasized: That's my personal opinion – and I'm no (official) "Collaborator" or "Contributor" here. I'm just an enthusiastic user who sometimes contributes to the Wiki – or publishes articles on GB to several places :smiley_cat:

walkjivefly:
Since this looks like becoming a bit of a discussion I'll add some thoughts.

[...]

PR #947, opened 2018-01-20
Changed target SDK version to 27 and fixed a few misspells

TaaviE:
All tests (except ignored FirmwareTest) passed. This also future-proofs GB against any Google Play restrictions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ android:
     - tools
 
     # The BuildTools version used by your project
-    - build-tools-26.0.2
+    - build-tools-27.0.3
 
     # The SDK version used to compile your project
-    - android-25
+    - android-27
 
     # Additional components
     - extra-android-m2repository
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,13 +17,13 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_7
         targetCompatibility JavaVersion.VERSION_1_7
     }
-    compileSdkVersion 25
-    buildToolsVersion '26.0.2'
+    compileSdkVersion 27
+    buildToolsVersion '27.0.3'
 
     defaultConfig {
         applicationId "nodomain.freeyourgadget.gadgetbridge"
         minSdkVersion 19
-        targetSdkVersion 25
+        targetSdkVersion 27
 
         // note: always bump BOTH versionCode and versionName!
         versionName "0.24.1"
@@ -64,13 +64,13 @@ dependencies {
     testCompile "org.robolectric:robolectric:3.5.1"
 
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:25.4.0'
-    compile 'com.android.support:cardview-v7:25.4.0'
-    compile 'com.android.support:recyclerview-v7:25.4.0'
-    compile 'com.android.support:support-v4:25.4.0'
-    compile 'com.android.support:gridlayout-v7:25.4.0'
-    compile 'com.android.support:design:25.4.0'
-    compile 'com.android.support:palette-v7:25.4.0'
+    compile 'com.android.support:appcompat-v7:27.0.2'
+    compile 'com.android.support:cardview-v7:27.0.2'
+    compile 'com.android.support:recyclerview-v7:27.0.2'
+    compile 'com.android.support:support-v4:27.0.2'
+    compile 'com.android.support:gridlayout-v7:27.0.2'
+    compile 'com.android.support:design:27.0.2'
+    compile 'com.android.support:palette-v7:27.0.2'
     compile('com.github.tony19:logback-android-classic:1.1.1-6') {
         exclude group: 'com.google.android', module: 'android'
     }
--- a/app/src/main/assets/app_config/js/gadgetbridge_boilerplate.js
+++ b/app/src/main/assets/app_config/js/gadgetbridge_boilerplate.js
@@ -85,21 +85,21 @@ function hideSteps() {
 function gbPebble() {
     this.configurationURL = null;
     this.configurationValues = null;
+
+    this.appMessageCallbackACK = {};
+    this.appMessageCallbackNACK = {};
+
     var self = this;
 
-    appMessageCallbackACK = {};
-    appMessageCallbackNACK = {};
-
-    function appMessageCallbackProcessed(transactionId) {
-    if (appMessageCallbackACK[transactionId]) {
-        self.removeEventListener("ACK"+transactionId, self.appMessageCallbackACK[transactionId]);
-        appMessageCallbackACK[transactionId] = undefined;
-    }
-    if (appMessageCallbackNACK[transactionId]) {
-        self.removeEventListener("NACK"+transactionId, self.appMessageCallbackNACK[transactionId]);
-        appMessageCallbackNACK[transactionId] = undefined;
-    }
-
+    self.appMessageCallbackProcessed = function (transactionId) {
+        if (self.appMessageCallbackACK[transactionId]) {
+            self.removeEventListener("ACK"+transactionId, self.appMessageCallbackACK[transactionId]);
+            self.appMessageCallbackACK[transactionId] = undefined;
+        }
+        if (self.appMessageCallbackNACK[transactionId]) {
+            self.removeEventListener("NACK"+transactionId, self.appMessageCallbackNACK[transactionId]);
+            self.appMessageCallbackNACK[transactionId] = undefined;
+        }
     }
     self.events = {};
     //events processing: see http://stackoverflow.com/questions/10978311/implementing-events-in-my-own-object
@@ -198,7 +198,7 @@ function gbPebble() {
                 }
                 var transactionId = GBjs.sendAppMessage(JSON.stringify(dict), needsTransaction);
                 if (needsTransaction) {
-                    if (callbackAck != undefined) {
+                    if (typeof callbackAck != "undefined") {
                         self.appMessageCallbackACK[transactionId] = function(e) {
 //                            console.log("ACK FOR " + JSON.stringify(e));
                             callbackAck(e);
@@ -208,7 +208,7 @@ function gbPebble() {
                         this.addEventListener("ACK"+transactionId, self.appMessageCallbackACK[transactionId]);
 
                     }
-                    if (callbackNack != undefined) {
+                    if (typeof callbackNack != "undefined") {
                         self.appMessageCallbackNACK[transactionId] = function(e) {
 //                            console.log("NACK FOR " + JSON.stringify(e));
                             callbackNack(e);
@@ -221,7 +221,7 @@ function gbPebble() {
             }
         }
         catch (e) {
-            GBjs.gbLog("sendAppMessage failed");
+            GBjs.gbLog("sendAppMessage failed: " + e);
         }
     }
 
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
@@ -172,7 +172,11 @@ public class GBApplication extends Application {
         if (isRunningMarshmallowOrLater()) {
             notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
             //the following will ensure the notification manager is kept alive
-            startService(new Intent(this, NotificationCollectorMonitorService.class));
+            if(!isRunningOreoOrLater()) {
+                startService(new Intent(this, NotificationCollectorMonitorService.class));
+            } else {
+                startForegroundService(new Intent(this, NotificationCollectorMonitorService.class));
+            }
         }
     }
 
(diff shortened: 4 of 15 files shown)


cpfeiffer:
In general, these changes are good, but did you actually test this on an Oreo device? Did you ensure that there are no further changes we have to do for claiming Oreo compatibility?

Because, once we use that API level, we can't go back.

TaaviE:
I tested it on my 7.0 device, I would have tested on a 8.1 emulator but there's a Linux kernel regression stopping me, reports from Google Play console indicate that there are issues on Android 8.1 so I've temporarily made it unavailable on those few devices.

Stacktrace:
```
android.app.RemoteServiceException: 
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1768)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:164)
  at android.app.ActivityThread.main (ActivityThread.java:6494)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:438)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:807) 
```

Issue exists on Google Pixel (sailfish) and Nexus 5X (bullhead). There seems to be a SO thread about the change https://stackoverflow.com/q/47531742 . If you see this as an easy fix and can do it then you should have access to the branch to push changes to it, otherwise I'll do it once my PC's kernel get's fixed.

TaaviE:
Related to the last commit, is there a reason "build pushed pull requests" is not enabled in Travis's settings?

cpfeiffer:
You mean that one? https://travis-ci.org/Freeyourgadget/Gadgetbridge/pull_requests

cpfeiffer:
Re the target version: sorry, but we don't intend to bump the target version just yet since we don't have devices for testing either. I never got the emulator to work properly and fully. Unless someone provides patches that are actually tested on a real Oreo device, we will stay with the current SDK level.

I'd advise to stay with that version as well. To put it mildly, it's not a nice thing to publish someone else's application on Play Store and patch it so that it crashes right on startup.

TaaviE:
@cpfeiffer The thing is that Google Play is going to require up-to-date target SDK with new releases, I'd rather get this out of the way rather now than when the deadline comes. I agree that crashing application isn't a good user experience, that's why I excluded those devices until your next release what I'll not patch like that.

About Travis, oh, it just doesn't update pull requests with the status then?

cpfeiffer:
re Travis -- I dunno, tbh.

IzzySoft:
@TaaviE you might blame me having a sharp tongue here – but maybe you start getting an idea for some of the reasons we'd rather stayed away from Playstore. You're only starting to face *some* of those issues now:

* you're forced to target a specific Android version for which you cannot even test yet
* you're limiting UX by having the app either crashing or devices excluded (I already hear the cries of "why" – and people thinking low of GB as their devices are not supported)
* you are forced to several other changes

And all that just to make the app "future-proof against any Google Play restrictions" it otherwise wouldn't even face.

I don't know if you do anyone a favor by that. Neither the project here, nor the users there. Nor do I see a good reason if you already *have to target Oreo* – especially as with Oreo, one can define F-Droid as a "secure source" (so that "unknown sources" thingy is no argument there).

Your wish to give GB a "wider base" is certainly laudable – but IMHO aimed in the wrong direction. I'm afraid that way might cause more harm than good.

OK, my 2 ¢ – and I'll shut up already.

TaaviE:
@IzzySoft I don't disagree that the restrictions are softly put, inconvenient, but targeting Oreo has to happen at one point anyways. I absolutely disagree that having a presence on Google Play doesn't have it's benefits, at least to users, you're also forgetting corporate environments where only Google Play is allowed. I agree that including this patch was not sufficiently tested and I'll try to avoid such changes in the future.

danielegobbetti:
@TaaviE you cannot expect to change the deadlines of an open source project like this. This is not how it works, this is not how any of this works.

Your actions are actively harming this project. If this is not what you want, then the sooner you realize this and step back, the less damage will be done. If this is what you want, you're doing a great job.

TaaviE:
@danielegobbetti I suggest you e-mail me about any concerns you have and explain a bit more what do you mean by "damage".

TaaviE:
> what was commonly agreed upon

What was commonly agreed upon by core devs (from https://github.com/Freeyourgadget/Gadgetbridge/issues/480) was:

> What we (core devs) don't want to happen is (malevolent) third parties distribute the code with the original package name that is altered in ways that could hinder the project (e.g. ads, other unwanted changes).

I respected those wishes fully, if that's not what you meant, then I'm really sorry, maybe you should have been more clear.

IzzySoft:
You could have asked before you've put it there. As for your quote: if that would be the final agreement, GB would have already been published e.g. by the OP of that issue, or by Avamander, who both offered that. So why didn't you quote e.g.

> But then again, I prefer a "blessed" play store release by the core devs than having a random guy publishing it.

Also written by a core dev, by the way.

TaaviE:
Tested on both API 26, 27 emulator and on API 19 physical device. Everything seems to work fine, also made sure the already-merged bugfix isn't causing any conflicts.

cpfeiffer:
Thanks, looks good now, but needs some thorough testing (always needed after a target version bump).
:+1:

TaaviE:
Small status update, more than ten days of the changes been published there are no API bump related crash reports (on more than 250 devices).

cpfeiffer:
That's great to hea, I hope we can merge this soon!

danielegobbetti:
I'm testing the PR on a oreo phone.

danielegobbetti:
I noticed some issues with a brief testing:
- quitting Gadgetbridge (from the menu) does not update the notification: according to the icon the device remains connected, same for the notification content. However logcat (and the watch) agree that the connection is indeed closed
  - the notification also shows the "disconnect" action button, that does nothing when selected.
  - tapping the notification launches the control center, after this is shown the notification icon updates and the situation seems back to normal
- after force-closing Gadgetbridge, the notification area shows a circle, and when swiping down there is a notification with title "Gadgetbridge" and content: "Gadgetbridge is running", "Touch to get more infromation or to stop background execution" (translated by me, might not be 100% correct). When tapping this notification the "app information" screen is shown. 

I will keep running this PR and add further observations here.

danielegobbetti:
Just noticed the following warnings:
```
nodomain.freeyourgadget.gadgetbridge W/Notification: Use of stream types is deprecated for operations other than volume control
nodomain.freeyourgadget.gadgetbridge W/Notification: See the documentation of setSound() for what to use instead with android.media.AudioAttributes to qualify your playback use case
```

TaaviE:
Thanks for testing this! 

> quitting Gadgetbridge (from the menu) does not update the notification: according to the icon the device remains connected, same for the notification content. However logcat (and the watch) agree that the connection is indeed closed

> after force-closing Gadgetbridge, the notification area shows a circle, and when swiping down there is a notification with title "Gadgetbridge" and content: "Gadgetbridge is running", "Touch to get more infromation or to stop background execution" (translated by me, might not be 100% correct). When tapping this notification the "app information" screen is shown.

Okay, I now see the issue on the emulator, it seems that GadgetBridge's process is not dying when it's told to quit. The issue seems to also exist pre-these-changes (at least on f03a3c1f3b936a82c7d6b068a31048b08b42dd46 it does) but is just invisible. I think this should be solved separately from this issue as it, as previously mentioned, exists on previous versions too.

Also just a clarification, the notification icon disappears, but later the circle appears, it seems that GB just restarts itself (improperly)? "Disconnect" doesn't work for previously mentioned reasons.

> nodomain.freeyourgadget.gadgetbridge W/Notification: Use of stream types is deprecated for operations other than volume control
nodomain.freeyourgadget.gadgetbridge W/Notification: See the documentation of setSound() for what to use instead with android.media.AudioAttributes to qualify your playback use case

I guess I'll update `setPriority()` to `setImportance()` to try and fix the warning, I can't see any other deprecated functions mentioned [here](https://developer.android.com/reference/android/app/Notification.Builder.html) being used. Though it's weird that they haven't annotated it deprecated (or Android Studio ignores it *shrug*).

TaaviE:
> Gadgetbridge cannot be really closed because of the notification listener, that much is true. The misbehavior of the icon/notification however is specific to this PR (might be some workaround is needed on oreo, not saying you're doing anything wrong).

I'm not sure how this can be worked around if GB can't be killed, as long as it's alive the notification will remain? When the listener is restarted the notification could be updated somehow, unless disabled in the settings the GB service should restart and update it's notification?

danielegobbetti:
> When the listener is restarted the notification could be updated somehow.
My main concern is exactly that according to the notification GB is running **and** connected. If it was updated to reflect the actual connection status everything would work correctly (Gadgetbridge would be relaunched upon tapping the notification, the "connect" action button would work).

I agree with you that the general aspect of the app not being killed is not part of this PR's discussion.

danielegobbetti:
I upgraded my phone to 8.1 and am now testing again the issues I reported earlier.

TaaviE:
I've been thinking a way to update the notification nicely from the notification listener and I can't think of a nice way. Maybe you have good ideas how to reliably do this?

danielegobbetti:
I believe I fixed the misbehavior, which I believe was introduced by the changes to ``NotificationCollectorMonitorService``.
1. creating the notification channel in there (I moved this to the end of GBApplication.onCreate())
1. adding a dummy notification there as well (I removed this code completely)

There are some other changes that I do not understand (like making the ``NotificationCollectorMonitorService`` a foreground service on Oreo) 

As said, I believe that my changes simplified the situation and everything appears to work correctly so far, but I will keep running it for some time to be sure. Can you please comment on the changes I mentioned as I don't understand the reason for them (but I understand that you blindly fixed a crash, and you couldn't test properly).

If you prefer I can add comments to the relevant lines.

TaaviE:
@danielegobbetti 

> There are some other changes that I do not understand (like making the NotificationCollectorMonitorService a foreground service on Oreo)

Background services aren't allowed on API26, there's a stacktrace about this in this PR too.

You might also want to push your changes because I can't see them.

danielegobbetti:
I just kept the (according to my tests) needed changes and squashed some of your commits on master (18cec2087f0081fd52c04e5cf85fb87f1be1e478 , with proper authorship preserver), and added my changes on top of it (e28333fe6e7e2279f5ed3a887171ee91e7ba22b5), mentioning your commit in the commit message.

The content of this PR is therefore merged on master, and I will close this. Thank you for motivating me into looking at API 26/27 compatibility.

TaaviE:
@danielegobbetti You forgot the foreground service change, the app crashes on Android 8.0 because background service launching from a background service is forbidden. 
> Unlike the ordinary startService(Intent), this method can be used at any time, regardless of whether the app hosting the service is in a foreground state.

Source: https://developer.android.com/reference/android/content/Context.html#startForegroundService(android.content.Intent)

> If your app needs to create a foreground service while the app is in the background, use the startForegroundService() method instead of startService().

Source: https://developer.android.com/about/versions/oreo/background.html

danielegobbetti:
The change was not forgotten, it just works fine as is here on android 8.1

danielegobbetti:
Just making the whole service a foreground one (like in the original PR) is disruptive for the user experience and it should be either limited to affected devices (or Android versions) or refactored so that the foreground service is a wrapper that terminates immediately after launching the background.

TaaviE:
@danielegobbetti 
> Just making the whole service a foreground one (like in the original PR) is disruptive for the user experience

I totally agree. 

> should be either limited to affected devices (or Android versions) 

The thing is that now there's a new issue that popped up, again startService, but from a different component:

```java
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:3273)
  at android.app.ActivityThread.-wrap17 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1689)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:172)
  at android.app.ActivityThread.main (ActivityThread.java:6637)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:767)
Caused by: java.lang.IllegalStateException: 
  at android.app.ContextImpl.startServiceCommon (ContextImpl.java:1505)
  at android.app.ContextImpl.startService (ContextImpl.java:1461)
  at android.content.ContextWrapper.startService (ContextWrapper.java:644)
  at ee.aegrel.gadgetbridge.impl.GBDeviceService.invokeService (GBDeviceService.java:84)
  at ee.aegrel.gadgetbridge.impl.GBDeviceService.disconnect (GBDeviceService.java:118)
  at ee.aegrel.gadgetbridge.externalevents.BluetoothStateChangeReceiver.onReceive (BluetoothStateChangeReceiver.java:53)
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:3266)
  at android.app.ActivityThread.-wrap17 (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1689)
  at android.os.Handler.dispatchMessage (Handler.java:105)
  at android.os.Looper.loop (Looper.java:172)
  at android.app.ActivityThread.main (ActivityThread.java:6637)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.Zygote$MethodAndArgsCaller.run (Zygote.java:240)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:767)
```

Devices affected:

| Device | Count |
|-|-|
|Mi A1 (tissot_sprout) | 7 |
|OnePlus5 (OnePlus5) | 7 |
|OnePlus5T (OnePlus5T) | 4 |
|OnePlus3T (OnePlus3T) | 2 |
|Xperia X Compact (F5321) | 1 |


Android versions:

|Android | Count|
|-- | -- |
|Android 8.0 | 20 |
|Android 8.1 | 1 |

PR #948, opened 2018-01-21
Improved notification text

TaaviE:
Added a small spacer that makes the notification a lot nicer to read, before the device name and the text were just concatenated without any space in-between.


cpfeiffer:
Thanks!

PR #949, opened 2018-01-21
Added a debug build configuration, enabled proguard for release

TaaviE:
Configured the debug build version of the app not to use the same package name and Pebble provider, the debug build configuration also provides a much easier way to test official and debug builds side-by-side without any hassle. In addition to that I set minifyEnabled and shrinkResources to true for release builds, they should be quite safe assuming no black magic has been done Proguard won't recognize.


TaaviE:
I'd also look into enabling proguard's optimizations, but that'd require more testing with more devices.

app/build.gradle
```diff
@@ -30,10 +30,20 @@ android {
...
+            shrinkResources true
+            proguardFiles getDefaultProguardFile("proguard-android.txt"),
+                    "proguard-rules.pro"
+            resValue "string", "pebble_content_provider", "com.getpebble.android.provider"
+        }
+
+        debug {
+            applicationIdSuffix ".debug"
+            versionNameSuffix "-debug"
+            resValue "string", "pebble_content_provider", "com.getpebble.android.debug.provider"
```
danielegobbetti: This change is equivalent to not having any content provider exposed in the debug build, as it's nowhere to be found in pebble official SDK (e.g.: https://github.com/pebble/pebble-android-sdk/blob/master/PebbleKit/PebbleKit/src/main/java/com/getpebble/android/kit/Constants.java ).

Practically this means that debug builds cannot interact with third party pebble apps, and this is clearly not an option.

TaaviE: Unfortunately the alternative is a content provider conflict. Which means there's no way to run release and debug on the same device. Those who wish to do so can still switch to building release and installing that on their device.

danielegobbetti:
-1 as this change brings dubious improvements in exchange for big confusion.

TaaviE:
Who's this going to confuse? I'm genuinely curious.

danielegobbetti:
Whoever used to run debug builds (which is the only possibility ATM) before this change and suddenly finds himself with a malfunctioning application.

NotAFile:
I disagree, I'd like this. And I don't use pebble so I don't mind that being broken. Perhaps there is a better way to do something equivalent?

TaaviE:
> Whoever used to run debug builds (which is the only possibility ATM) before this change and suddenly finds himself with a malfunctioning application.

If they used to run them they probably should understand how builds work, after all, it's also possible to use the release build to debug.

TaaviE:
Improved minifying, this means almost a 1MB size reduction without any loss of functionality. Also reenabled AAPT2 for safer and faster compilations.

TaaviE:
I wouldn't call the devices "problematic" just normal phones where one might want to test changes out and not replace any "production" data. As debugging the content provider is rare I don't see why it couldn't be automatically renamed to avoid any possible conflicts and if needed un-renamed.

matejdro:
I think there should be a way to explicitly enable this feature, but having it disabled by default (instead of having it another way).

Yes, this applies to "only Pebbles", but some people here only use Pebbles and don't care about other devices. You can't just say "it applies only to X" just because you don't use it.

TaaviE:
@matejdro This **only applies to debug builds** first of all and **only to the content provider** and **only if you have a pebble** the chances are super slim that's the case, and if you actually are you have the intelligence to manually disable it. Right now it's cumbersome to the majority.

matejdro:
Here is my proposal:

```
if (hasProperty('alternate_pebble_provider')) {
    resValue "string", "pebble_content_provider", "com.getpebble.android.debug.provider"
} else {
    resValue "string", "pebble_content_provider", "com.getpebble.android.provider"
}
```

Then in `.gradle/gradle.properties` in your user folder you can add `alternate_pebble_provider=true` line. Opt-in for everyone that needs it without the hassle for everyone else. Sounds like a win-win for me.

TaaviE:
> Then in .gradle/gradle.properties in your user folder you can add alternate_pebble_provider=true line. 

That's literally the same as it is right now - a hassle for everyone except that tiny minority who wants to debug the content provider and it also stupidly breaks release builds which should have the original content provider always.

matejdro:
Why does it break release builds? This line is only in `debug {}` block, release block stays the same.

I guess this comes down as a personal preference. I prefer to just having debug build on my device, running it as my main build to avoid hassle of switching between debug and release.

TaaviE:
> This line is only in debug {} block, release block stays the same.

You did not mention that in the original comment.

> I prefer to just having debug build on my device

I heavily disagree that debug builds as a release build should be an use-case catered to but not the people who want to effortlessly get debug and release running alongside.

matejdro:
Well, it seems we have reached a stalemate here. Both of us just have different workflow. It is probably best for project mantainers to decide on which solution is better.

PR #953, opened 2018-01-24
Fix crash when selecting automatic export interval

maufl:
I made a mistake in the code for the periodic export :(
Gadgetbridge crashes when you try to change the export interval. This should fix it. Sorry.


cpfeiffer:
Thanks for fixing!

PR #962, opened 2018-01-30
 fix privacy caller ID selection

criogenic:
fix privacy caller ID selection


cpfeiffer:
Thanks a lot for fixing this, nice catch :+1:

PR #972, opened 2018-02-01
Xwatch Basic Support

ladbsoft:
Basic support for the affordable Chinese Casio-like smart watches.

What does work:
* Activity syncing
* Multimedia control using the watch's button (one click play/pauses; double click skips to the next song/track)

What needs fixing:
* Intensity needs to be scaled down/up until a nice proportion
* Device must be set busy so the user cannot click sync again while syncing

What will be implemented next:
* Notifications
* Alarms

Will be fixing/implementing in my free time. Hope someone can make use of this :)

### DIFF
--- a/GBDaoGenerator/src/nodomain/freeyourgadget/gadgetbridge/daogen/GBDaoGenerator.java
+++ b/GBDaoGenerator/src/nodomain/freeyourgadget/gadgetbridge/daogen/GBDaoGenerator.java
@@ -65,6 +65,7 @@ public class GBDaoGenerator {
         addHPlusHealthActivityKindOverlay(schema, user, device);
         addHPlusHealthActivitySample(schema, user, device);
         addNo1F1ActivitySample(schema, user, device);
+        addXWatchActivitySample(schema, user, device);
 
         addCalendarSyncState(schema, device);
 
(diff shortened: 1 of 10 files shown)


ladbsoft:
Hi,

I have been trying to understand why the Travis CI build fails, but I couldn't. It's late in the night right now, I'll try to review it this weekend :)

NotAFile:
Nice! However, it seems in a few files you changed the formatting/line endings such that the whole file is marked as changed. This is unfortunate as it pollutes the history. Also, I would encourage you to use more descriptive individual commit messages in the future so changes can be better understood.

Either way, let's see if the maintainers mind.

ashimokawa:
@ladbsoft 
Wow, thanks! This is great.
However is is pretty much impossible to review the changes because you reformatted all lines in the files you touched. We use the default android studio style, if you reformat the files you changed again the diff should look much better.

About the commit messages, they are all the same, but I don't mind, I would squash this into one commit, it is only pressing a button in github, so don't worry.

NotAFile:
re: unauthorized:

This is might be because you have 2fa enabled. It is not very well documented, but pushing over https is not allowed with 2fa on.

cpfeiffer:
There's no UI for it, but you can append `?w=1` to the URL to ignore all whitespace changes, i.e. see this: https://github.com/Freeyourgadget/Gadgetbridge/pull/972/files?w=1

For travis, I also have no idea yet. Might be a problem on their side, or we somehow reached a limit in aapt.

In any case, nice job! :+1:

ashimokawa:
The recent conflict comes exactly from the problem with indention :/
Shall we clean the mess up or will you @ladbsoft 
Would be too bad to have this rotting

ladbsoft:
@ashimokawa: I'll try to redo the changes this afternoon, after work. Hopefully that will solve the problem. :)

ladbsoft:
Hi!

I solved the indentation problems, and the branches does not have any conflicts now. Travis build still fails... "You have not accepted the license agreements of the following SDK components: [Android SDK Platform 27].". It is right, I am testing the app in an oreo device, but I didn't change the target API :/

Any advice? Thanks!

ashimokawa:
It is not your fault. Travis is broken and we changed the target SDK.

Many thanks! Will have a look later today.

ashimokawa:
squashed and merged, thanks!

PR #997, opened 2018-02-16
Amazfit BIP and notification delay from incoming call

tomurbanowicz:
Is any chance for option of notification delay from incoming call? In MiFit can set delay of notification incoming call.


danielegobbetti:
**Please please please be careful!!!!!!!**. You opened a Pull request from another branch of this same project to master, asking a completely unrelated question in the opening comment.

Issue #989, opened 2018-02-12
Disconnection alerts


TaaviE:
While working on adding iTag/NUT support to Gadgetbridge I think this functionality could be implemented for all devices. What do you think?

danielegobbetti:
I believe that it makes sense to make it generic, but it could be impossible (read: very cumbersome) to distinguish an "out of range" event from a "device (BT) was turned off" one. 

There are also specific functionalities described by the Bluetooth SIG for this use case, but I guess that cheap key finders are not going to use those.

TaaviE:
> but it could be impossible (read: very cumbersome) to distinguish an "out of range" event from a "device (BT) was turned off" one.

I think it's not even necessary, in case where the user turns the device off I'm sure they can disable the alarm and after all it can also be left configurable.

> There are also specific functionalities described by the Bluetooth SIG for this use case, but I guess that cheap key finders are not going to use those.

Actually some do. Though that's link loss detection in another direction (wearable gives an alarm).

cpfeiffer:
We should also try to support the "almost out of range" scenario so that we can send a notification to the wearable before it's too late. Use case: someone steals your phone or you forget it somewhere. Would be great if your wearable would notify you.

Bonus points for integration with profiles, so that it won't bother you all the time e.g. at home.

TaaviE:
@cpfeiffer I thought about later implementing general "real time" alerts, be it heart rate, RSSI or some other parameter with optional alert destination.

cpfeiffer:
Yes, sounds good!

danielegobbetti:
potential duplicate / related prehistoric issue: #38

Issue #774, opened 2017-08-22
Feature Request: Support key finders


IzzySoft:
#### Before opening an issue please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
A new feature suggestion

#### Your wearable device is:
Pebble Time Steel

#### Your android version is:
5.1

#### Your Gadgetbridge version is:
0.19.4

----

There are several "key finders" available e.g. at Amazon, using BLE. They're usually 2-way (with a button), and the price range starts at [about EUR 5](https://www.amazon.de/dp/B01M2WF51S "Bluetooth anti-loss Tracker with carabiner") (unrated), having a good candidate at [about EUR 10](https://www.amazon.de/dp/B0734PWCYT "GOMAN Key Finder") (5 stars at 16 votes), another interesting one [at EUR 10](https://www.amazon.de/dp/B07259HV33 "Anti Lost Smart Key Finder belt-clip") (unrated), Two-for-One [at ~EUR 15](https://www.amazon.de/dp/B071FB1F7C "XinXu Key Finder") (5 stars at 12 votes) – and of course no definite "upper limit". All of them need a companion app, but interestingly multiple of them point to the very same ([iSearching](https://play.google.com/store/apps/details?id=com.lenzetech.antilost), with a lousy rating of 2.0 at 222 votes – though I bet [iKeeper_pro](https://play.google.com/store/apps/details?id=com.ds.ikeeper), [Smart Button nurum](https://play.google.com/store/apps/details?id=com.dothing.nurum) and others could be used as well) – which indicates a "common protocol". Behavior is quite obvious:

**In reach and connected:**

* press the button on the finder › phone rings an alarm
* press a button in the app › finder makes a noise
* use the finder as remote shutter for the camera, or to start a voice recording

**Connection is lost:**

* phone alarm rings

**Out of reach:**

* app shows "last seen time"
* for some, app remembers "last seen location" and can show it on a map

[...]

Issue #305, opened 2016-05-17
Multiple smart wearables at once?


ShapeShifter499:
I've noticed that I cannot sync or have two devices connected at the same exact time. Is this a bug? If not why can't two devices be connected at the same time?  My current use case is a Pebble Time smartwatch and a Mi Band 2 /w Heart Rate Sensor. I'd like to log data from both.

cpfeiffer:
At the moment, you can have only one device connected at a time. You could sync one device, disconnect and connect and sync the next device.

How do you want to correlate and combine the data?

ShapeShifter499:
At the moment I have no idea which device yields more accurate sleep and movement data, maybe someone with more use knows. I really would like to incorporate the heart rate monitor data from the MI band with everything else from the Pebble Time watch for graph data and especially if I choose the "live activity" mode. Is that possible?

Spleeeno:
I really hate the og pebbles vibration.
and even with fw 3.12 you can only enable or disable it.
so I'd love to use the pebbles display to read notifications and at the same time, get the (much smoother imho) vibration just on the miband (or amazfit).

although I'm not sure if thats even possible.
can my phones bluetooth chip even use normal bluetooth and le at the same time?
cause afaik pebble uses normal bluetooth and the mi band uses le.

anyway, maybe thats just me.
friggin vibration strength setting on the pebble would solve this too ;)

cpfeiffer:
@ShapeShifter499 Everything's possible with a little work ;-)

Getting and displaying the activity data simultaneously from two devices in realtime is one of the more challenging tasks, though, and we already have lots of other things to do. So we would need someone who wants to make this happen. Most developers either have a Pebble or a Mi Band, so even testing this is not that easy.

But as for all the other things: we're happy to assist!

cpfeiffer:
@Spleeeno I don't know whether it's possible to display notifications on the pebble without vibrating, but @ashimokawa or @danielegobbetti should know.

If that's possible, having the Mi Band vibrate and the Pebble display the notification should be possible. The bluetooth stack should support this scenario.

ashimokawa:
It is possible to turn off vibration in the settings menu on the pebble.

This task is not too easy since we assume one device at a time everywhere.
Me and @danielegobbetti have a Pebble and Mi Band.
And @cpfeiffer could use the emulator :speak_no_evil:

meta-meta:
[...]

PR #999, opened 2018-02-17
Added basic iTag support

TaaviE:
This commit should add basic "locate tag" functionality to Gadgetbridge. First step towards resolving [this issue](https://github.com/Freeyourgadget/Gadgetbridge/issues/774) This is so basic because I've got many issues to tackle before full functionality could be offered, namely:

- [ ] Support for multiple connections (lack of this is currently making link loss characteristic pretty impossible to use, https://github.com/Freeyourgadget/Gadgetbridge/issues/305 if you want more than one wearable)
- [ ] Easy way to rename devices (use of multiple trackers is rather cumbersome, for example I've got five of these tags and testing new changes is a headache when you've got "iTag", "ITAG", "iTAG" and etc. in the list :laughing:)
- [ ] Auto-discovery of characteristics and some manual exclusion wizard is needed because of two reasons:
    - [ ] Some devices have the ability to mute the beeping and just flash when instant alert is written, but not all of them, I can't think of a good way to detect this but giving the user the option would be nice (per-device)
    - [ ] Some devices advertise instant alert but actually don't beep when it's written (making locate unreliable as it might not exist in the tag)
- [ ] Easy way to set-up intents to trigger when the button on the device is pressed (with different tags)
- [ ] Live data alert support (https://github.com/Freeyourgadget/Gadgetbridge/issues/989)
- [ ] Nut* (Nut 3 mini for example) tags are much much more complicated, I haven't managed to make those beep even, the official app even contains mentions of DFU, will add support separately most likely


cpfeiffer:
Looks good so far :+1: Would be nice if you could use upper case first character for class names (`ITagCoordinator` instead of `iTagCoordinator`) as per Java naming conventions.

re renaming devices: do you mean on the db level or renaming the BT devices? If the latter, we need to make sure the device discovery based on BT name still works (not sure how persistent the name is).

re characteristics: you have all the available characteristics in the *Support class, so you could make use of them when available, or not use them otherwise. Or am I missing something?

re intent selection: yes, very much needed!

TaaviE:
@cpfeiffer 
> Would be nice if you could use upper case first character for class names (ITagCoordinator instead of iTagCoordinator) as per Java naming conventions.

Okay, can do.

> re renaming devices: do you mean on the db level or renaming the BT devices?

DB level

> re characteristics: you have all the available characteristics in the *Support class, so you could make use of them when available, or not use them otherwise. Or am I missing something?

That's the thing, some of these modules advertise some functionality but actually lack it (for ex. tags without buttons advertise they have a button). So I don't really know how setting those devices up could work properly.

cpfeiffer:
I see. Certain things need to be detected by hw version, fw version, appearance, ... See e.g. nodomain.freeyourgadget.gadgetbridge.service.btle.profiles.deviceinfo.DeviceInfo (deviceinfo profile: https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.service.device_information.xml).

So far we have custom solutions to store such additional information in the *support classes (see other DeviceInfo classes). Maybe we should make it possible to add such metadata to a GBDevice instance, like addDeviceInfo(ItemWithDetails), but a little more typed.

ashimokawa:
@TaaviE 

Will you do the changes requested by @cpfeiffer , maybe we can then merge for 0.25.0 release (the next one)

TaaviE:
Done. Though rest of the items on the TODOlist would have to be resolved for full support.

ashimokawa:
@TaaviE 
Yes, totally agree. Do you still want to have this merged now?

TaaviE:
I'm not opposed to it as it won't hurt anyone if it's there, but it doesn't have many practical uses either. It's up to you to decide.

TaaviE:
(Takes more work to resolve conflicts than the GH issue resolver provides, does not build)

Issue #482, opened 2016-12-31
Pebble weather for watchface


danielegobbetti:
The weather data is fetched by the means of [this app](https://f-droid.org/repository/browse/?fdfilter=gelin&fdid=ru.gelin.android.weather.notification), hence the Internet permission is **not** required and is not being added for this functionality to work.

The feature has been released in version 0.17 See the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki/Pebble-System-Apps#weather) to get instructions on how to enable this functionality. The supported watchfaces work on every pebble (including classic), the weather system app is only available on pebbles with firmware 4.

Currently requested watchfaces/apps and statuses:
- [x] System weather app
- [x] [timestyle pebble](https://apps.getpebble.com/en_US/application/55a5c024f4510f794c000071)
- [x] [Square](/turnervink/square)
- [x] [Healthyfy](/prakashait09/Healthify)
- [x] [Fancy](https://github.com/zalewszczak/pebble/tree/master/fancy)
- [x] Tally by [this user](https://github.com/zalewszczak/pebble) but sources not available?
- [x] Crowex by [this user](https://github.com/zalewszczak/pebble) but sources not available?
- [x] [Obsidian](https://github.com/stefanheule/obsidian)
- [ ] [TrekVolle](/aHcVolle/TrekVolle) partial support
- [ ] ~~[Timeboxed](/hussin/timeboxed-watchface)~~ crashes on pebble time
- [ ] [Simply Light](/bhdouglass/simply-light)
- [ ] [DigiMe](/workinghard/DigiMe)
- [ ] [forecal](/clreinki/forecal)
- [ ] [Eight Frame](/aemacleod/eight-frame)
- [ ] [AllThatInfo](/elieserleao/allthatinfo)

IzzySoft:
@danielegobbetti you could have copied our Conversation from yesterday for a starter :) Had I known you wanted it here, I had saved the links on my computer... So please forgive me when I ommit the links to the Pebble Store, and rather focus on links to the resp. Github projects:

* [Timeboxed](/hussin/timeboxed-watchface)
* [Simply Light](/bhdouglass/simply-light)
* [Square](/turnervink/square)
* **[TrekVolle](/aHcVolle/TrekVolle)**
* [DigiMe](/workinghard/DigiMe)
* [forecal](/clreinki/forecal)
* [Eight Frame](/aemacleod/eight-frame)
[...]

PR #1011, opened 2018-02-28
Pebble: add support for weather in Simply Light watchface

slp:
Celsius forced as with other watchfaces #482

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/AppMessageHandlerSimplyLight.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/AppMessageHandlerSimplyLight.java
@@ -1,0 +1,139 @@
+/*  Copyright (C) 2016-2018 Andreas Shimokawa, Sergio Lopez
+
+    This file is part of Gadgetbridge.
+
+    Gadgetbridge is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published
+    by the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Gadgetbridge is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>. */
+package nodomain.freeyourgadget.gadgetbridge.service.devices.pebble;
+
+import android.util.Pair;
+import android.widget.Toast;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.UUID;
+
+import nodomain.freeyourgadget.gadgetbridge.deviceevents.GBDeviceEvent;
+import nodomain.freeyourgadget.gadgetbridge.deviceevents.GBDeviceEventSendBytes;
+import nodomain.freeyourgadget.gadgetbridge.model.Weather;
+import nodomain.freeyourgadget.gadgetbridge.model.WeatherSpec;
+import nodomain.freeyourgadget.gadgetbridge.util.GB;
+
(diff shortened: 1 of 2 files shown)


danielegobbetti:
Nice! Thanks for the contribution.

I was wondering whether there is a key to make sure the watchface knows the temperature is in Celsius (see for instance [this line](https://github.com/Freeyourgadget/Gadgetbridge/blob/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/AppMessageHandlerTimeStylePebble.java#L140) )

slp:
@danielegobbetti 

> I was wondering whether there is a key to make sure the watchface knows the temperature is in Celsius (see for instance this line )

The conversion is done in the JS part:

simply-light/src/js/app.js:
`
258             var units = WeatherMan.KELVIN;
259             if (config.configuration.temperature_units == 'imperial') {
260                 units = WeatherMan.FAHRENHEIT;
261             }
262             else if (config.configuration.temperature_units == 'metric') {
263                 units = WeatherMan.CELCIUS;
264             }
265 
`
Unless there's a way we can read the configuration from Gadgetbridge and do the conversion beforehand, I don't it'd be possible to support choosing between units.

danielegobbetti:
Unrelated: does the watch face work with background js enabled?

slp:
> Unrelated: does the watch face work with background js enabled?

Nope. The problem appears to be related to some TypeErrors that start appearing after enabling background js. There are also some issues trying to generate the configuration data for the watchface. TBH, I haven't paid much attention to any of these issues.

danielegobbetti:
@slp can you please test if the errors are still there also in latest master? In our JSON message the metadata part was completely missing until latest commit.
Is the watchface open source? I would like to test it with background JS enabled.

I have no objections in merging this PR, but I would like @ashimokawa to give a second look. Anyway, +1 from here! 

Thanks again!

ashimokawa:
@slp 
Looks good to me. :+1: 

HOWEVER :stuck_out_tongue: 

We originally planned to remove all these handlers when background javascript works, but since that seems to be so complicated and you did the work already, I will still merge it. :wink:

IzzySoft:
@ashimokawa 

> We originally planned to remove all these handlers when background javascript works

… and does not consume twice as much battery. Still have this issue here. Additionally to the log I've sent you: I've noticed multiple reboots of my Pebble over the day, usually when some notification is received. Not on every notification, so I cannot reproduce it at-will …

Switching off BGJS, and those issues are not there.

Sorry for abusing this issue – but I jumped on "when BGJS works" :wink:

slp:
> @slp can you please test if the errors are still there also in latest master? In our JSON message the metadata part was completely missing until latest commit.
> Is the watchface open source? I would like to test it with background JS enabled.

@danielegobbetti Just gave it a try. It no longer complains about the missing transactionId, but the weather is still not showing up. I haven't looked further into this.

This watchface is GPLv3:

https://github.com/bhdouglass/simply-light

slp:
@ashimokawa 

> We originally planned to remove all these handlers when background javascript works, but since that seems to be so complicated and you did the work already, I will still merge it.

Thanks!

As @IzzySoft already commented, if it's not much of maintenance burden, having the option to use the handlers instead of the background JS would be a really nice option. IMHO, this is just another advantage of Gadgetbridge over the stock Pebble app.

danielegobbetti:
@slp the issue with the handlers is that the key IDs (potentially) change with version upgrades, which makes the whole process very fragile.

Before you ask: yes, that happened already with Timestyle, we updated the handler but of course doing so we fixed the new version by breaking the old one...

slp:
@danielegobbetti Yikes. On the other hand, is not possible to register the same handler with two different uuids? Just curious, I know that doesn't help with the maintenance burden.

danielegobbetti:
> On the other hand, is not possible to register the same handler with two different uuids?
Indeed it is possible, see for instance the ``UUID_ZALEWSZCZAK_*`` UUIDS, all handled by the same handler.

czandor:
Hello,

I want to make my watchface NineTime compatible with gadgetbridge. Now, the new unpublished version is working well, except the wind speed in current weather, and there is no forecast weather.
Please add wind direction and speed to mimicOpenWeatherMapResponse, and I need the forecast if possible.

IzzySoft:
@czandor see [here](https://github.com/Freeyourgadget/Gadgetbridge/issues/482#issuecomment-373929855) (same request from another watchface dev).

PR #1032, opened 2018-03-20
Update build.gradle

TaaviE:
First step towards migrating to Gradle 3.0 properly, changed all `compile`s to `implementation`s. Tests show only marginally faster compilation and no ill effects. https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html

Changed all `'` to `"` for consistency.

Also made a few whitespace and capitalization changes and removed an unnecessary `;`

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,14 +1,13 @@
+apply plugin: "com.android.application"
+apply plugin: "findbugs"
+apply plugin: "pmd"
 
-apply plugin: 'com.android.application'
-apply plugin: 'findbugs'
-apply plugin: 'pmd'
-
-def ABORT_ON_CHECK_FAILURE=false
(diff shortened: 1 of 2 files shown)


TaaviE:
It's also worth noting that there's a newer version of Android support libraries available `27.1.0`.

danielegobbetti:
Thanks, I will check other apps in the fdroid metadata repo to be sure gradle 3 is supported by them.

danielegobbetti:
Yeah sorry I miswrote, I meant checking for other apps using the new syntax.

danielegobbetti:
At least conversations is already using the new syntax, hence +1 from me. (To be complete, they are using single quotes but I can't imagine that creating problems).

PR #1035, opened 2018-03-22
added hr to gpx, adjust date format, adjust file nam

swiss5555:


ashimokawa:
@cpfeiffer 
Can you review this?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miband2/operations/FetchSportsDetailsOperation.java
```diff
@@ -97,7 +97,7 @@ protected void handleActivityFetchFinish(boolean success) {
             try {
                 ActivityTrack track = parser.parse(buffer.toByteArray());
                 ActivityTrackExporter exporter = createExporter();
-                String fileName = FileUtils.makeValidFileName("gadgetbridge-track-" + DateTimeUtils.formatIso8601(summary.getStartTime()));
+                String fileName = FileUtils.makeValidFileName("gadgetbridge-track-" + DateTimeUtils.formatIso8601(summary.getStartTime())+ ".gpx");
```
cpfeiffer: The idea is that we can have different kinds of exporters. The gpx-Exporter may create files with a `.gpx` extension, but other exporters may use a different extension. So we should add a method ActivityTrackExpoter#getExportFileExtension() and use its result here.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/export/GPXExporter.java
```diff
@@ -87,7 +88,7 @@ private void exportTrack(XmlSerializer ser, ActivityTrack track) throws IOExcept
         List<ActivityPoint> trackPoints = track.getTrackPoints();
         String source = getSource(track);
         for (ActivityPoint point : trackPoints) {
-            exportTrackPoint(ser, point, source);
+            exportTrackPoint(ser, point, source, trackPoints);
```
cpfeiffer: AFAICS this implements some kind of nearest-neighbour algorithm to find a valid heart rate value for those track points that lack one.

I would like to see this optimized a bit, since currently we would traverse the entire list of trackpoints again and again and again.

A simple optimization would be to assume that the samples are sorted by increasing timestamp. After finding the hr-sample for the first sample without hr, you would remember the index of the hr-sample and for the next sample you would start from there instead of from the beginning.

Additionally you could perform a custom binary search by the timestamp.

And last, I think there should be a maximum distance. If the nearest neighbour hr-value is 1h away, there's no point in using it. I think the maximum distance should be 1min or so.

swiss5555: Hi, If possible please do the optimization, I am purly c#/.net developer and for me is quite difficult to write code in java syntax :) not mention about android studio :) that why my code is a mess :)

ashimokawa:
@swiss5555 
There are now 21 commits in this PR, it should be rebased against master also (will delete bip-wip soon)

ashimokawa:
@swiss5555 
This got automatically closed since I deleted the branch. :/

PR #1036, opened 2018-03-22
Mi Band 2: wait at least 4s before sending notification text

mpiatka:
This fixes notification text not displaying when a short vibration
pattern is set.

If the notification text is sent while the icon is still visible
it is not displayed. We need to wait until it disapears (about 4 seconds)


ashimokawa:
@mpiatka 
Which firmware are you on? Just to reproduce here..

mpiatka:
Firmware: 1.0.1.69
Hardware: V0.1.3.3

Phone: HTC 10
Android 8

I am testing this via the debug menu with the GENERIC_SMS setting (the UNKNOWN setting never shows the message for me).

If I set the sms vibration pattern to Long and 3 repetitions the message shows after the vibration is over. If I set it to less repetitions or for example the Short pattern the mesage never shows.

ashimokawa:
@mpiatka 
Thanks, will look into it

PR #1039, opened 2018-03-22
WIP: Added multi-device support.

TaaviE:
This PR should add experimental multi-device support to Gadgetbridge, just in case I'll repeat, it's experimental. It works for me with three iTags (https://github.com/Freeyourgadget/Gadgetbridge/pull/999) quite well. I tried to solve the problem as described by cpfeiffer here: https://github.com/Freeyourgadget/Gadgetbridge/issues/305#issuecomment-269559313 , looking back this probably isn't the most elegant solution but the Bluetooth part has to be refactored later anyways.

Known bugs:
- [ ] - Giving multiple devices the "Locate" command causes the first iTag device to sometimes stop beeping (device-specific issue?)
- [ ] - Bluetooth state change listener doesn't do anything
- [ ] - Notification about Gadgetbridge's status needs to contain all devices(?)
- [ ] - Two toasts about devices being located won't/doesn't work, UI suggestions please
- [ ] - Database might get wrongly updated (? I have not managed to test it enough to prove otherwise, possible new test ideas?)
- [X] - Tests

I would really appreciate if someone had the chance to do test of these items:
- [ ] - Notifications routing (I only have one pebble that is able to display them)
- [ ] - Firmware updating
- [ ] - App management

Please do comment and suggest how to improve or what to fix in these changes :+1: and any help would also be really appreciated (maintainers should have direct push access).


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/profiles/deviceinfo/DeviceInfo.java
```diff
@@ -29,6 +29,7 @@
     private String systemId;
     private String regulatoryCertificationDataList;
     private String pnpId;
+    private String address;
```
cpfeiffer: I think this is not part of the spec, so it should be either avoided or marked at least.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceCommunicationService.java
```diff
@@ -282,36 +183,29 @@ public synchronized int onStartCommand(Intent intent, int flags, int startId) {
 
         Prefs prefs = getPrefs();
         switch (action) {
-            case ACTION_START:
+/*            case ACTION_START: {
                 start();
                 break;
```
cpfeiffer: Why did you comment this out?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceContainer.java
```diff
@@ -0,0 +1,201 @@
...
+    private TimeChangeReceiver timeChangeReceiver = null;
+    private BluetoothConnectReceiver blueToothConnectReceiver = null;
+    private BluetoothPairingRequestReceiver blueToothPairingRequestReceiver = null;
+    private AlarmClockReceiver alarmClockReceiver = null;
+
+    private AlarmReceiver alarmReceiver = null;
+    private CalendarReceiver calendarReceiver = null;
+    private CMWeatherReceiver CMWeatherReceiver = null;
+    private OmniJawsObserver omniJawsObserver = null;
+    private Random random = new Random();
```
cpfeiffer: Uh, do we really need to have one instance of all these per connected device?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/profiles/battery/BatteryInfo.java
```diff
@@ -19,9 +19,10 @@
 import android.os.Parcel;
 import android.os.Parcelable;
 
-public class BatteryInfo implements Parcelable{
+public class BatteryInfo implements Parcelable {
 
     private int percentCharged;
+    private String address;
```
cpfeiffer: I think this is not part of the spec, so it should be either avoided or marked at least.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceCommunicationService.java
TaaviE: Should've clarified (and will do) that the action is handled at line 151.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceContainer.java
TaaviE: If the receivers aren't configured per-device somehow then most likely not.

TaaviE: Though it will complicate the code quite a bit when one of the devices is disconnected then it has to be decided if the listener is needed any longer based on all the devices that are connected at that point?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/profiles/deviceinfo/DeviceInfo.java
TaaviE: How should I mark it? I would have avoided it if I would have seen a cleaner way of passing down for/from what device the message is meant for/coming from.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceContainer.java
TaaviE: All the receivers is one part of the code I haven't been able to verify working either, as mentioned I simply don't have multiple devices that support all the features.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothStateChangeReceiver.java
```diff
@@ -48,9 +48,9 @@ public void onReceive(Context context, Intent intent) {
                 }
 
                 LOG.info("Bluetooth turned on => connecting...");
-                GBApplication.deviceService().connect();
+                //GBApplication.deviceService().connect(); TODO: FIX ME!
```
TaaviE: What devices should it try to connect to?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBandPairingActivity.java
```diff
@@ -254,7 +254,7 @@ protected void performBluetoothPair(GBDeviceCandidate deviceCandidate) {
     }
 
     private void performApplicationLevelPair() {
-        GBApplication.deviceService().disconnect(); // just to make sure...
+        //GBApplication.deviceService().disconnect(); // just to make sure... //TODO: FIXME
```
TaaviE: I think this no device should be disconnected?

marcusfromadifferentage:
hi guys,

i want to say thank you for gadgetbridge in general and the progress you have made. 

i work in a hospital and we want to use amazfit bip devices to track patient activity. we want to keep the data inhouse, so gadgetbridge is very important for us. 

it would be wonderful if you could continue to implement the multiple device feature as this will allow us to work with more devices at the same time. 

so, again, thanks for you work.

marcus

Issue #1041, opened 2018-03-24
Feature Request: change FAB margin


Allen-B1:
#### Before opening an issue please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
a request. The [material guidelines](https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-keylines-spacing) say that there should be a 16dp margin.

Floating action button has different margins (10dp and 30dp): https://github.com/Freeyourgadget/Gadgetbridge/blob/41632b8eb3d76c46a1dcb5494be304228f85cefe/app/src/main/res/layout/activity_controlcenterv2_app_bar_main.xml#L36

#### Your wearable device is:
Pebble Classic

*Please specify model and firmware version if possible*

#### Your android version is:
7.0

#### Your Gadgetbridge version is:
0.25.0

*New issues about already solved/documented topics could be closed without further comments. Same for too generic or incomplete reports.*

danielegobbetti:
If anyone wants to start contributing to the project this would be a good first contribution!

PR #1042, opened 2018-03-26
Change fab margin

Allen-B1:
See issue #1041 

Also, removed elevation and pressedTranslationZ because `FloatingActionButton`s already have elevation


danielegobbetti:
will test

PR #1052, opened 2018-03-30
Fixed issue on P8 Lite and Power Four with BluetoothDevice getName() starting to return null

TaaviE:
Tries to fix this crash (has occurred twice, once on P8 Lite and once on Power Four:
```java
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:4805)
  at android.app.ActivityThread.access$1600 (ActivityThread.java:165)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1437)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:150)
  at android.app.ActivityThread.main (ActivityThread.java:5621)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:794)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:684)
Caused by: java.lang.NullPointerException: 
  at ee.aegrel.gadgetbridge.util.DeviceHelper.getBondedDevices (DeviceHelper.java:257)
  at ee.aegrel.gadgetbridge.util.DeviceHelper.getAvailableDevices (DeviceHelper.java:130)
  at ee.aegrel.gadgetbridge.devices.DeviceManager.refreshPairedDevices (DeviceManager.java:156)
  at ee.aegrel.gadgetbridge.devices.DeviceManager.<init> (DeviceManager.java:123)
  at ee.aegrel.gadgetbridge.GBApplication.onCreate (GBApplication.java:167)
  at android.app.Instrumentation.callApplicationOnCreate (Instrumentation.java:1015)
  at android.app.ActivityThread.handleBindApplication (ActivityThread.java:4793)
  at android.app.ActivityThread.access$1600 (ActivityThread.java:165)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1437)
  at android.os.Handler.dispatchMessage (Handler.java:102)
  at android.os.Looper.loop (Looper.java:150)
  at android.app.ActivityThread.main (ActivityThread.java:5621)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:794)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:684)
```

I considered it being Proguard's misdoing, but I heavily doubt it would get short-circuit evaluation wrong (I think I'd be seeing it a lot more then), I'm suspecting too frequent polling just sometimes returns null. .-.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/DeviceHelper.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/DeviceHelper.java
@@ -253,8 +253,11 @@ public class DeviceHelper {
         List<GBDevice> result = new ArrayList<>(pairedDevices.size());
         DeviceHelper deviceHelper = DeviceHelper.getInstance();
         for (BluetoothDevice pairedDevice : pairedDevices) {
-            if (pairedDevice.getName() != null && (pairedDevice.getName().startsWith("Pebble-LE ") || pairedDevice.getName().startsWith("Pebble Time LE "))) {
-                continue; // ignore LE Pebble (this is part of the main device now (volatileAddress)
+            String pairedDeviceName = pairedDevice.getName();
+            if (pairedDeviceName != null){
+                if(pairedDeviceName.startsWith("Pebble-LE ") || pairedDeviceName.startsWith("Pebble Time LE ")) {
+                    continue; // ignore LE Pebble (this is part of the main device now (volatileAddress)
+                }
             }
             GBDevice device = deviceHelper.toSupportedDevice(pairedDevice);
             if (device != null) {


ashimokawa:
This PR does nothing. It does not change behaviour at all.
In C like language if you have this:

`if (expression1 && expression2) {}`

then expression2 will only be evaluated if expression1 is true, so if expression1 checks for non null then expression2 can't do a null pointer dereference.

TaaviE:
@ashimokawa 
But every expression is a function call in very short succession? Like I said, I heavily doubt it's Proguard messing with short-circuit evaluation because then it wouldn't crash on only two devices.

cpfeiffer:
It's most probably just proguard or something else messing up the line numbers in the stacktrace. We had that already in the mpandroidchart crashes, remember?

TaaviE:
@cpfeiffer 
Assuming it is proguard messing with the code (and line numbers), there aren't many places there could be a NPE:

The function isn't being called with a null

```java
        if (btAdapter != null) {
            List<GBDevice> bondedDevices = getBondedDevices(btAdapter);
            availableDevices.addAll(bondedDevices);
        }
```

so the first line can't be the cause. `pairedDevices` could be null when 

> If Bluetooth state is not STATE_ON, this API will return an empty set.

but then just `result` is with a zero initial capacity and `pairedDevices` foreach doesn't get executed meaning there is no chance of a NPE. Assuming calling `.getName()` three times in a row isn't the issue then the only possibility is `GBDevice device = deviceHelper.toSupportedDevice(pairedDevice);` but that would mean `deviceHelper` is `null` which can't be the case?

Thus I can't see that any other line than `if (pairedDevice.getName() != null && (pairedDevice.getName().startsWith("Pebble-LE ") || pairedDevice.getName().startsWith("Pebble Time LE "))) {` and a faulty `getName()` implementation the actual cause of the issue. 

The original function for reference:

```java
    private List<GBDevice> getBondedDevices(BluetoothAdapter btAdapter) {
        Set<BluetoothDevice> pairedDevices = btAdapter.getBondedDevices();
        List<GBDevice> result = new ArrayList<>(pairedDevices.size());
        DeviceHelper deviceHelper = DeviceHelper.getInstance();
        for (BluetoothDevice pairedDevice : pairedDevices) {
            if (pairedDevice.getName() != null && (pairedDevice.getName().startsWith("Pebble-LE ") || pairedDevice.getName().startsWith("Pebble Time LE "))) {
                continue; // ignore LE Pebble (this is part of the main device now (volatileAddress)
            }
            GBDevice device = deviceHelper.toSupportedDevice(pairedDevice);
            if (device != null) {
                result.add(device);
            }
        }
        return result;
```

cpfeiffer:
The docs of `getBondedDevices()` also say
> Set<BluetoothDevice> | unmodifiable set of BluetoothDevice, or null on error

so I suppose that's where the NPE came from. I also added a check for null entries in the set, just to be sure.

TaaviE:
@cpfeiffer 
Thanks! Let's hope it fixes the crash.

PR #1057, opened 2018-04-06
Add setting to reset last sync date

andrewheadricke:
Adds a simple option for Amazfit BIP users to reset the last sync date in the Gadgetbridge Database.  That way its possible to re-download older activities.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/SettingsActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/SettingsActivity.java
@@ -23,6 +23,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
+import android.content.SharedPreferences;
 import android.database.Cursor;
 import android.location.Criteria;
 import android.location.Location;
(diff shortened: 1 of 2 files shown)


danielegobbetti:
Thanks for the contribution, but I am afraid it cannot be merged as is:
- settings never depend on a specific device, in this case you need the device to get the HW address that is used to create the preference key ==> I believe the preference should be changed somewhere else, not in the settings activity
- reconstructing the preference key (as you did correctly) is prone to errors in case of refactoring.

All in all, I agree this is a needed functionality, but I am afraid it must be done differently. In any case I leave to @cpfeiffer the last word.

Thanks again for the contribution in any case! But I must say -1 for the reasons stated above.

andrewheadricke:
No worries, i fully expected that it would need to be tweaked, but thought it was a good place to start.

Yeah, the activities summary overflow menu would be ok.  

My original code had the reset button in the debug activity, but I thought the BIP specific settings was a better long term solution.. I can change the PR to use the Debug Activity if you like, but I prefer the overflow suggestion :)

danielegobbetti:
Proposal: after marking an activity summary for deletion and pushing delete, we could show a modal asking whether the user wants to reset the ts so that the activity will be downloaded again from the watch.

andrewheadricke:
what happens if they click "don't reset ts" and then later on decide they do want to reset? Then we'd still need a button :)  Besides, users wanting this action would be pretty rare I imagine.  An obscure button somewhere for those rare times would suffice. just my 2cents.

danielegobbetti:
In my local uncommitted code I have a menu entry (in the overflow menu of the activity) that triggers a datepicker widget and resets the ts to that value. I was just thinking of something quicker to trigger :-)

danielegobbetti:
This has been solved in the activities activity, still reconstructing the key though. Thanks again for the contribution!

PR #1101, opened 2018-05-20
Added support for Q8 devices

tiparega:
Hello,

I have a Q8 device that is another HPlus. They all start with a Q8... name, so I added it as a new HPlus Subdevice. I tested it and seems to work well (as well as HPlus can).

Regards!
Alex


dgm3333:
Thanks Alex,
Sadly my watch died after a couple of weeks. I've ordered another, but it's a different style so I won't be able to test your patch. Hope yours lasts much longer.
David

ashimokawa:
Thanks ;)

PR #1126, opened 2018-06-15
Add RomanPort's AppStore Endpoint

KonradIT:
Adds `https://pebble-appstore.romanport.com/watchapps/0/?query={query}` as an Appstore Endpoint.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/appmanager/AbstractAppManagerFragment.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/appmanager/AbstractAppManagerFragment.java
@@ -428,7 +428,7 @@ public abstract class AbstractAppManagerFragment extends Fragment {
                 startActivity(startIntent);
                 return true;
             case R.id.appmanager_app_openinstore:
-                String url = "https://apps.getpebble.com/en_US/search/" + ((selectedApp.getType() == GBDeviceApp.Type.WATCHFACE) ? "watchfaces" : "watchapps") + "/1?query=" + selectedApp.getName() + "&dev_settings=true";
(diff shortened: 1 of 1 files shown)


IzzySoft:
Can anyone open that page in a browser? I wanted to check, but for me it doesn't work. Tried IE on a Windows machine, it says not supported, I should use Firefox. Using Firefox, all I get is that orange Night-Rider style bar on top, but no content.

KonradIT:
@IzzySoft [works here](https://i.imgur.com/AEVJB1k.png) https://pebble-appstore.romanport.com/watchapps/0/?query=dialer

IzzySoft:
@KonradIT What browser are you using? All I get with that URL is the same as above, just a flashing orange bar. And yes, I've enabled JS and cannot see anything being blocked in uBlock either. But I see a ton of errors in the console, like "Cross-Origin Request Blocked" (it obviously loads JQuery from some Google-URL, but has "Same Origin" in the policy), lots of "unreachable code after return statement" and so on.

Looks like the page code was adapted from the official Pebble store, but at least that one is working for me. I just tried it in Chrome (as that's installed on the machine I'm currently on; I usually do NOT use it due to Google tracking), there it strangely works …

It also loads tons of other stuff from Google servers (like fonts). So it's rather a privacy invasion.

Ah, just found what was blocking it: DecentralEyes side-loaded JQuery. Obviously that was either incompatible, or causing the policy violation.

Sorry for the confusion I might have caused! But that side should really do something for privacy. They could provide all that sideloaded stuff locally, for example. No problem with locally hosting the fonts and JavaScript. Unfortunately, the "contact" link just points to another privacy invader (Discord), so I won't report it. Wait, they have a [Github presence](https://github.com/Roman-Port/Pebble-Appstore). Will file an issue there.

ashimokawa:
@KonradIT 
Thanks, will merge just before pebble store is closing.

danielegobbetti:
Thanks for the contribution. I'm merging this now and we will tag a release just before the 30th.

KonradIT:
Rebble.io store (apps.rebble.io) should be good to go, if you want that one instead.

Issue #1116, opened 2018-06-03
Mi Band 2 - Automatic sync of activity data [Feature Request]


kaefert:
#### Before opening an issue please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
I wished there was the option for an automatic sync of activity data from the wearable into the Gadgetbridge Database on the phone.

Sometimes I don't open the Gadgetbridge app for a few days and I would like to not have to press the Sync button manually (or wipe down on the statistics page) to get my activity data from my Mi Band 2 into the Gadgetbridge Database.

Otherwise I'm quite happy with Gadgetbridge and particularly with the notification features and configuration options! 

#### Your wearable device is:
Mi Band 2
FW: 1.0.1.53
HW: V0.1.3.3

#### Your android version is:
LG G6, Android 7

#### Your Gadgetbridge version is:
0.26.5

danielegobbetti:
The main problem with this, is that any notification that arrives while the wearable is busy will not be forwarded to it, which is not what the user is expecting.

kaefert:
Well, maybe you could make the time for auto sync configurable so that it's possible to make it happen during the night when most people have silenced their notifications anyway.

danielegobbetti:
I was thinking of starting a fetch when the screen gets turned on, but apparently we can't get that event :(

ritualz:
I'm been missing this feature too, and kinda implemented it. It will basically fetch the activity on screen-unlock.

There is a few additional things I believe is needed:
1. time-interval between fetches (so that it not fetch the data every time the device is unlocked), where can i find this info?
2. check if the device are connected and busy. The DeviceManager inside the new  BroadcastReceiver, report false to these, even when the device is connected.


https://github.com/ritualz/Gadgetbridge/commit/da82c0ce056af863e35030feb9cd26bb82642b8d

danielegobbetti:
@ritualz can you make a pull request with your changes?

LuccoJ:
[...]

PR #1127, opened 2018-06-18
Auto sync feature

ritualz:
#1116, requested a feature which makes the application able to automatically refresh the activity data without the need to open the application. I think this might be a possible solution, where the activity data is transferred when the screen is unlocked. There might be a possibility for improvement, where the snackbar tooltip could be removed, and a period between the syncs (since this will always update, when the phone is unlocked). 

I've been using this patch for a few weeks with the MiBand2 and it seems to work without major problems.


danielegobbetti:
> seems to work without major problems.

are there minor problems to report? :-)

ritualz:
Not bugs, but the snackbar is annoying, and should be removed (in my opinion). But i could not find the function to sync the activity without it. 

screenshot_20180619-173017

danielegobbetti:
The screenshot contains a toast, or do you mean the snackbar in controlcenter?

ritualz:
I'm taking about the toast (not that into the whole android dev stuff). Is there a way to get rid of it?

danielegobbetti:
I will move the toast to a notification after having merged this. Thanks!

danielegobbetti:
please test current master, as the toast is now gone :)

ritualz:
Yeah, the toast is gone, thanks :+1:

PR #1137, opened 2018-06-30
Allow the user to decide the interval between autofetches

ritualz:
So this is a follow up from #1127, with a added option to block fetches within a time-period. 

The phone can be unlocked frequently, which would lead to GadgetBridge to fetch the activity data every time. This patch will disable the fetch for a period of time (set by the user). Also discovered that there was multiple GBAutoFetchReceiver active for one device, which should be removed now.


ashimokawa:
Thanks!

Issue #913, opened 2017-12-20
[feature request] Keep continuous heartrate on while connected on MiBand2


Dododappere:
After turning continuous heart rate on, it turns itself off after 15 seconds. The turning off can be prevented by writing 0x16 to the HRControlPoint characteristic after receiving a HR measurement. That way HR stays on for as long as the device is connected...

ashimokawa:
@Dododappere 
Oh cool, maybe we can remove our crappy timer code then in live activity also

silberzwiebel:
I'm completely new to gadgets and also to gadgetbridge. Thanks for your impressive work so far!

A clarification question, as I couldn't resolve this from reading some issues here (if I overlooked something: sorry for the noise): Is continuous heart rate measuring for the mi band 2 working in the latest release or is it turning off after 15 seconds? If so, would the solution from @Dododappere enable a fully working continuous measurement (given the band is connected to the phone)?

ashimokawa:
@silberzwiebel 
You can enable periodic HR measurements (every 10 minutes for example), this also works with the device not connected!

The Amazfit Bip (and maybe also the Mi Band 2, I didn't test) with this option enabled switch to continuous HR measurements when being active (running, etc), this means more that once every 10 minutes.

@Dododappere 
You mean we still have to write 0x16 every time after an HR sample? Where do you think this would make sense? In real time activity data (last tab in the graphs)? Or while being normally connected?

boun:
@ashimokawa from a quick test with mi fit it seems like it sends 0x16 every 10 seconds. I'll test it on my contentprovider branch this weekend where it would make total sense.

![image](https://user-images.githubusercontent.com/3662452/42936591-c27c7e8a-8b4c-11e8-9886-46edecee9884.png)

boun:
Hi, it works perfectly. POC Code would be this commit: https://github.com/boun/Gadgetbridge/commit/e22d72e8b5aa73cc4dd4e9f339ce2dcbbdeabdca

boun:
Some more info: Since got the contentprovider (https://github.com/Freeyourgadget/Gadgetbridge/pull/1138) working i was not really convinced by the reported values from my mi band 2. I first searched for a bug in my code, compared the adb logs, then I found this report and developed the POC Code.

The difference is striking: First screenshot is my typical running routine, and code similar to what I found in LiveActivityFramgment (disable/enable HR reporting each second). 

The HR Rate data makes me look like a pro:

[...]

PR #1138, opened 2018-07-04
[WIP][RFC] Introduce a ContentProvider for HR Data

boun:
Hi,

as I really want to get HR Data into runnerup here is a patch to do so. It is still not finished however I am proposing at an early stage to get some feedback. The Idea is to provide three query apis to:

- Discover Device(s?) The Interface looks like you can pair multiple devices. Hence an endpoint to discover them. As I only have a MiBand2 I am not 100% sure how Gadgetbridge behaves here
- Start / Stop Realtime DataCollection: It seems like I have to send an Intent to start that
- Get Realtime Data: Send Samples and inform ContentObservers

I have written some tests. What works for sure is discovery of a device. Realtime data is currently delivered as String, that is also a TODO. Furthermore my activity_start / stop are a stub in the tests.

However before implementing Contract Classes, deciding on the data format, writing more tests etc. I would like to have some feedback, if the API meets your "taste", what Data you would send (only Heartrate, Heartrate and Steps, Full Sample Object). Furthermore some input on the Code, how to test everything is highly appreciated. If you think ContentProvider is not the way to go, that is also fine.

Have fun!


boun:
About the Travis error: The testsuite does not set up the DeviceManager and I have no idea how to do that. Strangely locally all tests pass.

app/src/main/AndroidManifest.xml
```diff
@@ -413,6 +413,11 @@
             android:authorities="com.getpebble.android.provider"
             android:exported="true" />
 
+        <provider
+            android:name=".contentprovider.HRContentProvider"
+            android:authorities="com.gadgetbridge.heartrate.provider"
```
cpfeiffer: gadgetbridge.com is not our domain, gadgetbridge.org is. But why do you explicitly not use the nodomain.freeyourgadget.org name?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -0,0 +1,238 @@
+/*  Copyright (C) 2016-2018 Andreas Shimokawa, Carsten Pfeiffer
```
cpfeiffer: This should be your name, here :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -0,0 +1,238 @@
...
+        Log.i(HRContentProvider.class.getName(), "Creating...");
+        IntentFilter filterLocal = new IntentFilter();
+
+        filterLocal.addAction(GBDevice.ACTION_DEVICE_CHANGED);
+        filterLocal.addAction(DeviceService.ACTION_REALTIME_SAMPLES);
+
+        LocalBroadcastManager.getInstance(this.getContext()).registerReceiver(mReceiver, filterLocal);
+
+        //TODO Do i need only for testing or also in production?
+        this.getContext().registerReceiver(mReceiver, filterLocal);
```
cpfeiffer: Gadgetbridge's own events are sent via LocalBroadcastManager, so this *should* not be necessary.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProviderContract.java
```diff
@@ -0,0 +1,21 @@
...
+
+import android.net.Uri;
+
+public final class HRContentProviderContract {
+    public static final String[] deviceColumnNames = new String[]{"Name", "Model", "Address"};
+    public static final String[] activityColumnNames = new String[]{"Status", "Message"};
+    public static final String[] realtimeColumnNames = new String[]{"Status", "Heartrate", "Steps", "Battery"};
+
+    static final String AUTHORITY = "com.gadgetbridge.heartrate.provider";
+
```
cpfeiffer: com.gadgetbridge again..

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/database/SampleProviderTest.java
```diff
@@ -1,32 +1,59 @@
...
         super.setUp();
         dummyGBDevice = createDummyGDevice("00:00:00:00:10");
+        mContentResolver = app.getContentResolver();
+
+        HRContentProvider provider = new HRContentProvider();
+        // Stuff context into provider
+        provider.attachInfo(app.getApplicationContext(), null);
+
+        ShadowContentResolver.registerProviderInternal("com.gadgetbridge.heartrate.provider", provider);
     }
```
cpfeiffer: com.gadgetbr...

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -0,0 +1,238 @@
...
+    // Returns the requested device. If it is not found
+    // it tries to return the "current" device (if i understand it correctly)
+    @Nullable
+    private GBDevice getDevice(String deviceAddress) {
+        DeviceManager deviceManager;
+        deviceManager = ((GBApplication) (this.getContext())).getDeviceManager();
+        for (GBDevice device : deviceManager.getDevices()) {
+            if (deviceAddress.equals(device.getAddress())) {
+                Log.i(HRContentProvider.class.getName(), String.format("Found device device %s", device));
+                return device;
```
cpfeiffer: Please use the slf4j logger instead of Log.i, Log, e, etc.

app/src/main/AndroidManifest.xml
boun: Indeed, I wanted to follow the name of the java packages, however on a second thought this should then have been **nodomain.freeyourgadget.gadgetbridge**. I'll change it to org.gadgetbridge. Because I a am tempted to keep the provider more general i'll go with

`org.gadgetbridge.realtimesamples.provider`

Feel free to oppose

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
boun: Either way, for testing I need that hook. Maybe it is because I did not understand the testing framework completely. I'll have a look. I think there is a `GBApplication.isLocal()` call, so I can call it only if a testcase is running.

boun:
So I addressed your comments for the better part. What is left now to do is:

- After some measurements I do not get any more heartrates from my Mi2BandSupport. This is unfortunate. It may have several reasons including: A bug in my code, a cursed firmware, a thread inside of GB which just shuts that reporting down. Any thoughts / pointers here? Could someone with another device test? Do I need to take a Wakelock somewhere?
- Fix Travis: Tests fail because I try to query the device Manager which is not set up correctly (at least getDeviceManager() in GBApplication returns null) in the test environment. I am open to suggestions for workarounds.
- Permission System to read HR Data: I think android asking for permission instead of just send HR Date to every interested party would be a nice thing to do

app/src/main/AndroidManifest.xml
```diff
@@ -415,7 +415,7 @@
 
         <provider
             android:name=".contentprovider.HRContentProvider"
-            android:authorities="com.gadgetbridge.heartrate.provider"
+            android:authorities="org.gadgetbridge.realtimesamples.provider"
```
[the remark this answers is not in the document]
boun: Then I misunderstood you. I'll finally change it to nodomain.freeyourgadget.gadgetbridge

boun:
O.k. I finally found the problem with the tests:

https://github.com/Freeyourgadget/Gadgetbridge/blob/11c545362ab4597dcf7333fd4d2982a264984e69/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java#L148

GBApplication is not set up correctly in onCreate(). It checks wether a static variable is set already and if so it returns. Rationale is to guard against multiple invocations from robolectric. However the DeviceManager is **not** static, therefore each invocation of the testsuite with **more than one** testcase will get `null` as a result for `((GBApplication)getContext()).getDeviceManager()` in the **second** and following testcases. 

I think I'll open up a separate issue.

cpfeiffer:
re timer: yes, we have to re-enable the reatime measurements again and again for Mi2. It is also kind of sensible, because it delivering realtime data is much more battery intensive than sending the data in one go. So Mi2 only sends it for a short period of time if it doesn't get another enable-command.

cpfeiffer:
re DeviceManager: back then we got multiple invocations of GBApplication#onCreate() by robolectric, either due to a bug or by design. Things may have changed since then, so we might revise this. I totally agree that this code is a little convoluted, with the db being setup either by GBApplication (in normal mode) or by TestBase (in test mode). Initially, robolectric did not support custom Application subclasses at all and I don't know how it handles the Application instance lifecycle at the moment. For db tests, we want to make sure that every test gets a fresh db environment.

According to your findings, I suspect that we do get a new GBApplication instance (hence deviceManager field being null). The static fields survive, of course.

So to sum up,
- we can check if we still get multiple onCreate() calls on the same instance
- we can try to make all static instances in GBApplication non-static and then remove the guard
- or we could make the deviceManager instance also static (urgh)
- or come up with something else?

ashimokawa:
@cpfeiffer 
I think we can have realtime measurements without sending commands over and over again. Not 100% sure though

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/huami/miband2/MiBand2Support.java
```diff
@@ -758,6 +758,7 @@ public void onHeartRateTest() {
         }
     }
 
+    // Unser Einstig
```
cpfeiffer: If these comments are intended to be submitted, they should be in English :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/huami/miband2/MiBand2Support.java
```diff
@@ -1193,6 +1197,7 @@ public void doCurrentSample() {
                     try (DBHandler handler = GBApplication.acquireDB()) {
                         DaoSession session = handler.getDaoSession();
 
+                        // Why is this so complicated????
```
cpfeiffer: Do you have an idea how to make it less complicateD?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -140,85 +138,130 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
...
+                // have to enable it again and again to keep it measureing
+                GBApplication.deviceService().onEnableRealtimeHeartRateMeasurement(true);
+            }
+
+        }, 1000 * 10, 1000);
+        // Start after 10 seconds, repeat each second
+    }
+
+    @NonNull
+    private Cursor realtime(String[] projection, String[] args) {
```
cpfeiffer: From looking at the name and the arguments, I would have no idea what it is or what arguments it expects. Maybe name it getRealtimeHeartRateCursor()?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -140,85 +138,130 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
...
-                    mc.addRow(new Object[]{"OK", buffered_sample.getHeartRate(), buffered_sample.getSteps(), mGBDevice != null ? mGBDevice.getBatteryLevel() : 99});
-                return mc;
+                LOG.info("REALTIME");
+                return realtime(projection, selectionArgs);
         }
         return null;
     }
 
+    @Nullable
+    private Cursor devices_list() {
```
cpfeiffer: Suggestion for a better method name (and please try to use camel case): getDevicesCursor()? Or just getDevices()?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -140,85 +138,130 @@ public Cursor query(@NonNull Uri uri, String[] projection, String selection, Str
...
+        GBApplication.deviceService().onEnableRealtimeHeartRateMeasurement(false);
+        mc = new MatrixCursor(HRContentProviderContract.activityColumnNames);
+        mc.addRow(new String[]{"OK", "No error"});
+        punchTimer.cancel();
+        punchTimer = new Timer();
+        return mc;
+    }
+
+
+    private void enable_realtime() {
```
cpfeiffer: enableContinuousRealtimeHeartRateMeasurement()? Long, but descriptive.

cpfeiffer:
Thanks for your work, really looking forward to getting this in! :+1:

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/huami/miband2/MiBand2Support.java
boun: Sorry, I managed to merge the [DO NOT MERGE] commit

boun:
About the tests: I did not see multiple invocations of onCreate during my (limited) tests. I saw one invocation per testcase. Hence from that point of view I should be o.k. to remove the guard. 

Currently I am playing around with the integration of the contentprovider into runnerup, https://github.com/boun/runnerup/commits/contentprovider and some minor cleanup. When I am done I'd rebase, open up a new pull request and squash everything together? 

Future Ideas for this provider would be:
- Add a library via maven which contains the ContractClass for easier integration
- Change the ContentProvider from this quite rudimentary `buffered_sample` in Memory thing to use the database as every realtime measurement automatically lands there too
- Provide historic data

boun:
After removing the guard **all** 61 tests still pass. onCreate is called 61 times.

boun:
And to answer that question: Yes this starts gadgetbridge in case it was not running before.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/contentprovider/HRContentProvider.java
```diff
@@ -0,0 +1,288 @@
...
+                    if (mGBDevice.isConnected() && state == provider_state.CONNECTING) {
+                        LOG.debug("Device connected now, enabling realtime " + mGBDevice);
+
+                        enableContinuousRealtimeHeartRateMeasurement();
+                    }
+                    break;
+                case DeviceService.ACTION_REALTIME_SAMPLES:
+                    ActivitySample tmp_sample = (ActivitySample) intent.getSerializableExtra(DeviceService.EXTRA_REALTIME_SAMPLE);
+                    //LOG.debug("Got new Sample " + tmp_sample.getHeartRate());
+                    if (tmp_sample.getHeartRate() == -1)
```
cpfeiffer: Better use HeartRateUtils.isValidHeartRateValue() here.

boun:
Sound lovely, I'll address the comment and squash it to a bare minimum. I'll ping you when I am done. In the meantime, might I motivate you to compile my runnerup fork and try it out with different / multiple devices? You don't need to go running, just some 5 mins of testing in your pocket would be fantastic. 

Furthermore, for the record: as soon as this change is merged, you have an external api, which others will develop against. Hence I'd flag it as "beta" or "unstable" in the changelog entry for the first version.

TaaviE:
I have one suggestion though, this content provider should be protected by an authorization dialogue before being released to the public and doing it right now while nothing depends on it sounds like a reasonable idea:

1. Fragment for managing content provider accesses  
2. Database or shared preferences model for storing package names and granted access(es, if we add more providers later on)
3. Content provider has to check calling package (For ex. `ContentProvider.getCallingPackage`)

boun:
Hi @TaaviE, thx for your comment. I think your wish is actually an OS function based on the android permissions system (see screenshot, the default is off, which I changed). Together with the suggestion by cpfeiffer (the big todo) it is guaranteed that the user first has to give his consent an then also sees an indicator that realtime monitoring is happening. And this will be the topic of my next comment :)

screenshot_1532211117

boun:
Hi @cpfeiffer, I like the idea of showing a notification to the user and allowing her to interrupt realtime sampling. However that is actually not as easy to do, because there is a thread inside the contentprovider (and LiveActivityFragment but that should not be the problem), that restarts that sampling each second. 

Coincidently I do not like this implementation. And even more of a coincident is, that for my mi band 2 it is probably not correct either (https://github.com/Freeyourgadget/Gadgetbridge/issues/913). Hence should this not be moved up the stack closer to the "drivers"?

TaaviE:
@boun Gadgetbridge's minimum API is Kitkat (19, 4.4) and runtime permissions are a Marshmallow (23+) thing meaning that this has to be implemented by Gadgetbridge.

TaaviE:
@boun Exactly my point, everyone who wants gets access when there's just that permission on KK. Say some proprietary app actually adds Gadgetbridge support I and I suspect many others would not want it to instantly have access to my heart rate data.

boun:
I consider your point highly hypothetical :) However feel free to implement it.

danielegobbetti:
Given the nature of the project, I agree that the feature outlined by @TaaviE is important and should be present when the new functionality is available in the **published** application.

Which means (in my opinion) that it could either be implemented in the PR, or after the merge (but before we release).

cpfeiffer:
I'm in a middle position. IMHO it shouldn't be a blocker for the first release. AFAIU, this is only a real problem for KK, where you cannot deny permissions. On Lollipop and above, a user could install a "rogue" app and simple deny the single permission. 

A simple solution might be to have this feature on Lollipop+ only. Not sure if that would be a problem for the RunnerUp userbase. but I could imagine they wouldn't want to require a Gadgetbridge permission anyway, if they still targeted KK.

I personally had not the possibility to try out the integration yet, maybe next week.

cpfeiffer:
What's the current state of this? Can we move this forward?

lesensei:
Hey there,
Since I'm very interested in being able to track my HR in an opensource running app, I tried this with my miband3. I probably did something wrong, because GB loses the connection to my mb3 after about 5 minutes and crashes when I try to export the logs from the debug screen. @boun if you're still interested in working on this, I can help test it but I'm no code wizard so I won't be able to do much by myself except run ADB and find out where it fails. I have runnerup from your contentprovider branch crashing as well when trying to get into the history pane, but I don't think this has anything to do with your changes.
Anyway, if I can help with something that's not way over my head, please let me know, I'd be happy to.
Regards,

cpfeiffer:
@lesensei Is this reproduceable? Can you provide logs (privately, if you want)?

PR #1143, opened 2018-07-09
Add MyKronoz Zetime support

Lightwars:
Hi, 

please see my changes to add some basic support for the MyKronoz ZeTime, which involves:
- sending notifications to the watch
- syncing activity data to the database: steps, calories, distance, time of activity, heart rate and sleep data
- sending weather information to the watch
- sync the time on the watch with your device
- control music playback from the watch

I am no java developer so I might have done some quirks.
Commits until Jul 06, 2018 are relevant.


cpfeiffer:
Nice :-)
We can't merge this as is, since there are unrelated commits from different people in this PR. Did you cherry-pick them? Travis failed because the executable bit was lost on the gradlew script. Did you perhaps develop on Windows, which does not understand the executable bit?

Do you think you can fix up your branch so that only your commits will be new or do you need help with that? You can force-push the branch without a problem, if you want.

PR #1148, opened 2018-07-10
Add MyKronoz Zetime support

Lightwars:
Hi,

please see my changes to add some basic support for the MyKronoz ZeTime, which involves:

  - sending notifications to the watch
  - syncing activity data to the database: steps, calories, distance, time of activity, heart rate and sleep data
  - sending weather information to the watch
  - sync the time on the watch with your device
  - control music playback from the watch

This time the commits are correct.


ashimokawa:
@Lightwars 
Nice!
@danielegobbetti @cpfeiffer 
Any objections merging this as is?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
```diff
@@ -0,0 +1,1223 @@
...
+                    break;
+            }
+            handleGBDeviceEvent(musicCmd);
+        } else {
+            if (music != null) {
+                music[2] = ZeTimeConstants.CMD_REQUEST_RESPOND;
+                try {
+                    TransactionBuilder builder = performInitialized("replyMusicState");
+                    replyMsgToWatch(builder, music);
+                    performConnected(builder.getTransaction());
```
cpfeiffer: I don't understand the pattern
```
performInitialized()
doStuff()
performConnected()
```

Probably the performXX() methods lack information, so here's an explalation:
- at first, the device is becoming connected (BT-level)
- most devices usually need some initialization like setting the time, etc. after connection
- after that, the device is 'initialized'
- performInitialized() will allow you to write to the device in a way that will make sure that the writes will only happen *after* the connection to the device has been established *and* the initialization has successfully finished
- afterwards, you need to actually perform the writes by calling `builder.queue(getQueue())`. This will return immediately -- the rest will happen asynchronously via the queue.

So the `performConnected()` call should typically not be necessary.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimeSampleProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public int toRawActivityKind(int activityKind) {
+        return activityKind;
+    }
+
+    @Override
+    public float normalizeIntensity(int rawIntensity) {
+        return rawIntensity;
```
cpfeiffer: Unfortunately not specified by the javadoc, but this method should return values between `0` and `1`. So you will probably need some calculation here.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
```diff
@@ -0,0 +1,1223 @@
...
+
+    }
+
+    @Override
+    public void onDeleteNotification(int id) {
+
+    }
+
+    @Override
+    public void onNotification(NotificationSpec notificationSpec) {
```
cpfeiffer: Would love to see this huge method being split into smaller ones :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
```diff
@@ -0,0 +1,1223 @@
...
+        int subject_length = 0;
+        int notification_length = 0;
+        byte[] subject = null;
+        byte[] notification = null;
+        Calendar time = GregorianCalendar.getInstance();
+        byte[] datetimeBytes = new byte[]{
+                (byte) ((time.get(Calendar.YEAR) / 1000) + '0'),
+                (byte) (((time.get(Calendar.YEAR) / 100)%10) + '0'),
+                (byte) (((time.get(Calendar.YEAR) / 10)%10) + '0'),
+                (byte) ((time.get(Calendar.YEAR)%10) + '0'),
```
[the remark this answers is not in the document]
Lightwars: The watch want to have ascii characters for some time stamps, so I convert the numerics to ascii numbers. Like 0 equals '0' to 9 equals '9'. I do not know if there is a java function already for this.

Lightwars:
@cpfeiffer @ashimokawa @danielegobbetti 
https://github.com/Freeyourgadget/Gadgetbridge/pull/1148/commits/ed20f69cc411d605076ecb356f790e26e548efc9

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
Lightwars: Had seen this pattern been used by other devices, but it might be that this was old or not right.

cpfeiffer:
Regarding the handleGBDeviceEvent() method -- usually you don't call it directly. Instead you call evaluateGBDeviceEvent(event), which delegates to the specific handle* method depending on the event type.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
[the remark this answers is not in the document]
cpfeiffer: I see, can you add a comment to the code somewhere so that this knowledge will be preserved? :-)

cpfeiffer:
Great work. When you think you're ready to merge, just say so. You can continue to change and fix this afterwards.

ashimokawa:
@Lightwars 
I am also ready to merge this, but be aware that this will then released to public soon because there are also other changes in master that I want to have released ;)

Lightwars:
For the time being this ready.
Will try to improve and enhance the support in the future.
:)

Lightwars:
@cpfeiffer I could do that.

cpfeiffer:
:+1:

Issue #1012, opened 2018-03-03
request for device ck11


Ilia1:
#### Before opening an issue please confirm the following:
- [ *] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [ *] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [ *] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
Device not visible

#### Your wearable device is:

bracelet ck11, with display, work with app Wearfit (official) and Wearfit2.0 and nRF Connect (connect, no data)

#### Your android version is:
Android 6
#### Your Gadgetbridge version is:
Frome fdroid 0.24.6

Ilia1:
Datasheet [nRF51822_PS_v3.1.pdf](https://github.com/Freeyourgadget/Gadgetbridge/files/1782008/nRF51822_PS_v3.1.pdf)

Ilia1:
https://github.com/SunnyLy/SmartBracelet?files=1
this app work too, but on android > 6

Ilia1:
https://github.com/NordicSemiconductor?tab=repositories
repo NordicSemiconductor, developer app nrf connect

Ilia1:
[Log 2018-06-13 16:33:38.txt](https://githucom/Freeyourgadget/Gadgetbridge/files/2097916/Log.2018-06-13.16.33.38.txt)
Log from nrf connect when connecting to wearfit

Ilia1:
in the bracelet is used softdevice 110, what version is unknown.
http://www.nordicsemi.com/eng/node_176/ANT/nRF51422#softdevice
in the archive there are header files for working with services

PR #1153, opened 2018-07-13
ID115 support

Disasm:
Here is some support code for ID115 fitness tracker from Aliexpress.

Technical info:
* CPU: nRF51822
* Protocol: Bluetooth LE
* Android app link: https://play.google.com/store/apps/details?id=com.veryfit2hr.second

Implemented features:
* Time sync
* Notifications
* Step goal setting
* Fetching activity data

Missing features:
* Activating sleep tracking
* Fetching sleep data

Also I need some help with infrastructure things. Such as:
* Incorrect icon shown on the main screen (pairing screen shows correct one).
* Notifications don't work when received from apps, but do work from Debug screen.
* How should I implement time-consuming things like health stats synchronization? Is there any API for that?


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/id115/ID115Coordinator.java
```diff
@@ -0,0 +1,135 @@
...
+import nodomain.freeyourgadget.gadgetbridge.model.ActivitySample;
+import nodomain.freeyourgadget.gadgetbridge.model.DeviceType;
+
+public class ID115Coordinator extends AbstractDeviceCoordinator {
+    @NonNull
+    @Override
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public Collection<? extends ScanFilter> createBLEScanFilters() {
+        ParcelUuid service = new ParcelUuid(ID115Constants.UUID_SERVICE_ID115);
+        ScanFilter filter = new ScanFilter.Builder().setServiceUuid(service).build();
```
cpfeiffer: Weee, discovery *not* based on the device name :-) :ok_hand:

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/id115/ID115Coordinator.java
```diff
@@ -0,0 +1,135 @@
...
+    public DeviceType getSupportedType(GBDeviceCandidate candidate) {
+        if (candidate.supportsService(ID115Constants.UUID_SERVICE_ID115)) {
+            return DeviceType.ID115;
+        }
+        return DeviceType.UNKNOWN;
+    }
+
+    @Override
+    public int getBondingStyle(GBDevice deviceCandidate){
+        return BONDING_STYLE_NONE;
```
cpfeiffer: Does this only work without pairing?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/id115/ID115Support.java
```diff
@@ -0,0 +1,479 @@
...
+    }
+
+    @Override
+    public void onSetTime() {
+        try {
+            TransactionBuilder builder = performInitialized("time");
+            setTime(builder);
+            performConnected(builder.getTransaction());
+        } catch(IOException e) {
+        }
```
cpfeiffer: Please don't swallow exception, log them at least.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/id115/ID115Support.java
```diff
@@ -0,0 +1,479 @@
...
+        try {
+            getQueue().clear();
+
+            TransactionBuilder builder = performInitialized("reboot");
+            builder.write(normalWriteCharacteristic, new byte[] {
+                    ID115Constants.CMD_ID_DEVICE_RESTART, ID115Constants.CMD_KEY_REBOOT
+            });
+            performConnected(builder.getTransaction());
+        } catch(Exception e) {
+        }
```
cpfeiffer: Same here.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/id115/ID115Support.java
```diff
@@ -0,0 +1,479 @@
...
+            outputStream.write((byte)currentNotificationSize);
+            outputStream.write((byte)currentNotificationIndex);
+            outputStream.write(raw);
+            byte cmd[] = outputStream.toByteArray();
+
+            TransactionBuilder builder = performInitialized("notification");
+            builder.write(normalWriteCharacteristic, cmd);
+            performConnected(builder.getTransaction());
+        } catch (IOException e) {
+        }
```
cpfeiffer: and here

cpfeiffer:
> Incorrect icon shown on the main screen (pairing screen shows correct one).
That's weird. Which icon does it show?

> Notifications don't work when received from apps, but do work from Debug screen.
Do you have notification access enabled for Gadgetbridge? Otherwise it won't be able to receive the notifications.

> How should I implement time-consuming things like health stats synchronization? Is there any API for that?
All BT communication is done in a background thread, so the UI thread will never be blocked. You can update notifications to show progress. Unfortunately this kind of synchronization is always very device specific and a little complicated. You can have a look at the FetchActivityOperation classes, for example.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/id115/ID115Coordinator.java
Disasm: Unfortunately, yes. I tried to bond with GadgetBridge and nRF Connect, but without success. So I added this code to skip bonding.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/id115/ID115Support.java
Disasm: I will fix that.

Disasm:
> That's weird. Which icon does it show?

[Disconnected icon](http://forever.disasm.info/files/github/Screenshot_20180715-000938.png)
[Connected icon](http://forever.disasm.info/files/github/Screenshot_20180715-001048.png)
[Correct icon](http://forever.disasm.info/files/github/Screenshot_20180715-001356.png)

> Do you have notification access enabled for Gadgetbridge? Otherwise it won't be able to receive the notifications.

Sure. Looks like there is a bug in master branch of GadgetBridge or I am missing something important. Notifications don't work with Pebble too. Privacy guard is disabled for GadgetBridge.

> You can have a look at the FetchActivityOperation classes, for example.

I will look, thanks. I have reversed sync mechanism today and it seems to be simple. Host sends sync command and then tracker sends multiple responses ending with some kind of "end of sync" response. I don't know what will happen if I will send another request during the sync, so it will be better to avoid such situations somehow. Also there are two sources for sync: steps and sleeping stats. Maybe the second sync should start after the end of the first.

cpfeiffer:
Those icons are... surprising. Something somewhere is terribly wrong ;-)

Regarding notifications, it might help to reboot your phone. There is, or was a bug in Android that prevents notifications to be delivered to your app if you app once crashed. No toggling of BT, restarting the app or anything helps. Rebooting always fixed it, though.

See the "BUSY" state of GBDevice -- it will prevent any other events from being delivered to your *Support instance until you clear the busy flag.

Disasm:
> Those icons are... surprising. Something somewhere is terribly wrong ;-)

I even tried to drop application data via system settings. Doesn't help.

Also, another strange thing. My device was not in list on the main screen after the search until I added non-null string to ID115Coordinator::getManufacturer(). Despite this, connect succeeds and notification says "connected". But no device on the main screen. Such relation between manufacturer string and main screen contents is far from obvious. Maybe it should be noted somewhere of fixed if it is a bug.

> Regarding notifications, it might help to reboot your phone. There is, or was a bug in Android that prevents notifications to be delivered to your app if you app once crashed. No toggling of BT, restarting the app or anything helps. Rebooting always fixed it, though.

Wow. This is so weird. Rebooting really helps, thank you!

ashimokawa:
@Disasm 
For the icons to show correctly, you have to edit level_list_device.xml, please merge in master in your branch before because we had ZeTime support just merged

Disasm:
@ashimokawa, thanks!

Disasm:
I have some questions about activity information gathering:
* Is GBDaoGenerator.java the only place where I should declare my activity sample information?
* Should I declare SAMPLE_RAW_INTENSITY property and set fake value for it if I have no information about activity intensity? Actually, I can approximate intensity as activeTime/totalTime, but I doubt that such approximation is useful.
* Should I create zero-activity records if there is no activity for some time intervals (zero step count reported by fitness tracker)?
* Apart from step count, ID115 reports some approximated values such as distance travelled and calories burned. Should I store them too or step count alone will be sufficient?

cpfeiffer:
1. Yes, GBDaoGenerator.java is the only place. Did it generate entity classes for you? You can use them in your custom `SampleProvider` implementation, then.

2. No, please don't try to calculate an intensity value. Just leave it out.

3. No, your samples need not be in the same interval. Just create those samples that you get and make sure they have a proper timestamp.

4. Those are probably not reported in an activity sample, but in some kind of summary, right? I've wanted to have support for such daily summaries or workout summaries, but never got to it. So atm, there is no place in Gadgetbridge, that would display this information. You could use `GBDevice#addDeviceInfo()`, then they could be displayed, but that would only be a poor workaround for a start. Essence is: you can record them, and you could also add support for such summaries, if you want :-)

Disasm:
Thank you for the answers!

1.Yes, GBDaoGenerator generates a class for my description. Seems like magic for me, so I asked just to be sure that I do the right thing.

4.This fitness tracker reports summary for the whole day, but it also reports 15-minute samples with the same items (steps, calories, distance travelled and activity time). I wonder what to do with the last three from the sample data.

Disasm:
Finally, I have done with activity data fetching. Current device support code reflects almost everything I have reverse-engineered so far. Further improvements require additional experiments with official app because a lot of features (like alarms and sleep tracking) should work, but I can't activate them with raw commands. Maybe they are not implemented in firmware at all.

cpfeiffer:
:+1: :+1: Great work!

Ilia1:
https://github.com/Freeyourgadget/Gadgetbridge/issues/1012
My device too on the processor nrf 51822
Will it work with the application and what data do I need to get that would make it work?

Disasm:
According to your log, your device has different Bluetooth LE haracteristics than ID115. It also has Nordic UART service (which is absent in ID115), so I can assume that these two devices are very different despite having the same MCU.

PR #1177, opened 2018-07-27
Add missing break + Allow devices to not support being found

joserebelo:
Not all devices should have the "find device" button.

Also, there's a missing break; in the createBTDeviceSupport switch.


cpfeiffer:
Good catch, thanks! :+1:

PR #1178, opened 2018-07-28
Splash screen and minor UI tweaks

joserebelo:
This PR adds a simple splash screen (with no artificial delay or overhead), preventing the white screen while the activity loads. It currently uses the primary colors and the Gadgetbridge image (which could use a new higher resolution or vectorized version), but that could change, I'm open to suggestions.

splash

It also tweaks the device icon top margin a bit, so it's vertically centered in relation to the labels:

card


TaaviE:
Awesome! 

I also agree that gadgetbridge_img.png should really be replaced by an svg. Though I couldn't quickly find the original of that image (or is the one commited the original?). Maybe the launcher icon .svg could be modified instead to be suitable?

joserebelo:
I think the one committed in #278 is the original. I wonder if @bytehala has a better quality source, else I could try to make a vectorized version

joserebelo:
I have updated gadgetbridge_img with a higher dpi version (not pixel-perfect, but I think it's faithful enough).

I have also tweaked the colors a bit, since in the dark theme the image was way too bright.

danielegobbetti:
Looks very good. Thanks. 

I am just looking around if I manage to find an SVG of the background before merging.

joserebelo:
I have replaced the image with a white version, which should fix the splash on API19. Since the tint still works on ImageView elements, no further changes should be needed.

I can also squash the last 2 commits if you want.

danielegobbetti:
I managed to get it to work with vector drawables. I'll push later to your branch.

Good news is that with vector API 19 shows no splash screen, hence everything just works (TM)

joserebelo:
@danielegobbetti but doesn't that introduce a delay while the activity loads, with no indication that it's loading on API 19?

When I tested the splash with vector drawable the splash wouldn't show on any API, but neither would the white screen. So nothing would show up at all until the activity finished loading.

danielegobbetti:
Well, I omitted that I created a splash file for android < 21 that does only contain the background. That is tested and works on API 19.

I cherry picked your first commit of this PR and just made the changes for using the vector resource and the drawable-less splash screen in a second commit. This way we don't add to the history the multiple PNGs. I hope you don't mind.

Please check out master to see if everything is there. If everything is fine we can close this.

Thanks a lot!

joserebelo:
Yes, it's fine by me.

This way, though, on API < 21 it will just show up an orange screen. That is always better than having a white screen or the black logo on orange, though.

danielegobbetti:
Sorry I did not cherry pick ca9229c ! Will do this later.

Issue #382, opened 2016-08-07
Heart rate - dealing with strange values


ghost:
From time to time the device seems to return the "doubled" heart rate. This might be an error in the measurement or an extra heartbeat which happens ... anyone else seeing this?
This can be relatively safely recognised by comparing current measurement with previous and next ones

Instead of plotting the wrong value it might be better to plot the half value with a special marker showing that something was unusual about this measurement.

IzzySoft:
Might look a bit harsh to simply close this – but it has been 3 years, and OP has even already removed his/her Github account so no feedback is possible. Should someone else still have this problem, please open a new issue. I assume this has already been solved, e.g. with the linked MR.

Issue #381, opened 2016-08-07
Live activity should not plot invalid heart rate


rgh36167:
it appears my miband returns "0" for about half the hr measurements and those are plotted resulting in a wild zigzag line going up and down from zero to valid value and back

In java/nodomain/freeyourgadget/gadgetbridge/activities/charts/LiveActivityFragment.java line 163 is a check for isValidHeartRateValue(heartRate) which causes setCurrentHeartRate(); not to be called.

However, in  getCurrentHeartRate() mHeartRate was previously set to "-1" to signal an invalid value and is returned by next calls. 
Where getCurrentHeartRate is callled in add entries there is no test for isValidHeartRateValue resulting in adding and plotting the bogus values.

Either in line 219 there should be a check for isValidHeartRateValue or as a quick hack getCurrentHeartRate() should remember and return the latest valid value to bridge over invalid measurements.

cpfeiffer:
Totally true, plotting heart rate values needs a lot of improvements.

uwehermann:
Confirmed, this also (unsurprisingly) happens with Mi Band 2 (HW V0.1.3.2, FW V1.0.0.34), GB 0.14.3.

cpfeiffer:
Mabye we need a discussion about the desired hr plotting.

IMHO we should be as accurate as possible, which means, we should *not interpolate* and create the impression that there are successful measurements with the same heart rate all the time.

I agree that those ups and downs look pretty bad, though, so maybe we should find some kind of compromise.

One possibility would be to interpolate, but paint in a different colour, so that one can recognize that there are no actual measurements. At some point, when the next valid measurement is too far away, we should probably drop interpolation between these values. Any suggestion for a sensible time span?

uwehermann:
Well, I would certainly suggest to not use the invalid heart rates at all, to avoid the "zigzag" for sure.

Other than that there are multiple options to show a nice graph, e.g. a "dots and lines" approach which shows the graph as "lines" overall, but there are small "dots" in the lines which indicate where an actual sample was received/used.

Example:
http://www.gnuplotting.org/figs/non-continuous_lines.png

As for interpolation / smoothing etc. there are various ways to calculate a "current heartrate" number as well in case you want to show that. Some "proper" ECGs use simple averaging of e.g. the last 6-8 beats or such, others use something like "Take the last 8 beats, drop the higest and lowest value, average the remaining 6". There are other methods as well.

[...]

PR #1181, opened 2018-07-29
remove incorrect HR data <10 from the graphs

### LINKED ISSUE #1173: Mi Band 3 indicates heart rate "3" a lot of the time
screenshot_20180726-001541

I know it's only preliminary support, but can't hurt to drop the issue I guess since it's the first I find with this device. The firmware is 1.2.0.8, hardware version 0.18.3.4, and Gadgetbridge is 0.28.1.

Setting a periodic heart rate reading (I tried every minute and every 5) results in many or most of the reading being "3", while the others are plausible (and fairly accurate compared to my Bip, too, I suppose).

I think that maybe "3" is just an error code indicating it failed to take that particular reading reliably, but I'm obviously guessing.

Dikay900:
Straightforward setting the minimum for the hr data to 10 instead of 0 to definitely eliminate zeroes and out of line hr data. Maybe we could even increase the minimum to around 40? What do you think?
Maximum value seems good to me.

<details>
  <summary>Before / After</summary>
before:
screenshot_20180801-205250

after:
screenshot_20180801-205639
</details>

fixes #1173 

Would like to let #382 and #381 open since there could be more bogus stuff around according to the discussions.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
@@ -18,7 +18,7 @@ package nodomain.freeyourgadget.gadgetbridge.activities;
 
 public class HeartRateUtils {
     public static final int MAX_HEART_RATE_VALUE = 250;
-    public static final int MIN_HEART_RATE_VALUE = 0;
+    public static final int MIN_HEART_RATE_VALUE = 10;
     /**
      * The maxiumum gap between two hr measurements in which
      * we interpolate between the measurements. Otherwise, two
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
@@ -36,15 +36,10 @@ public class HeartRateUtils {
     private int maxHeartRateValue;
     private int minHeartRateValue;
 
-    /**
-     * extra class used for synchronized access of the singleton
-     */
-    private static final class HeartRateUtilsInstanceHolder {
-        static final HeartRateUtils INSTANCE = new HeartRateUtils();
-    }
+    private static final HeartRateUtils instance = new HeartRateUtils();
 
     public static HeartRateUtils getInstance() {
-        return HeartRateUtilsInstanceHolder.INSTANCE;
+        return instance;
     }
 
     /**
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
@@ -16,9 +16,14 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>. */
 package nodomain.freeyourgadget.gadgetbridge.activities;
 
+import nodomain.freeyourgadget.gadgetbridge.GBApplication;
+import nodomain.freeyourgadget.gadgetbridge.util.GBPrefs;
+import nodomain.freeyourgadget.gadgetbridge.util.Prefs;
+
 public class HeartRateUtils {
     public static final int MAX_HEART_RATE_VALUE = 250;
     public static final int MIN_HEART_RATE_VALUE = 10;
+
     /**
      * The maxiumum gap between two hr measurements in which
      * we interpolate between the measurements. Otherwise, two
@@ -28,7 +33,42 @@ public class HeartRateUtils {
      */
     public static final int MAX_HR_MEASUREMENTS_GAP_MINUTES = 10;
 
-    public static boolean isValidHeartRateValue(int value) {
-        return value > HeartRateUtils.MIN_HEART_RATE_VALUE && value < HeartRateUtils.MAX_HEART_RATE_VALUE;
+    private int maxHeartRateValue;
+    private int minHeartRateValue;
+
+    /**
+     * extra class used for synchronized access of the singleton
+     */
+    private static final class HeartRateUtilsInstanceHolder {
+        static final HeartRateUtils INSTANCE = new HeartRateUtils();
+    }
+
+    public static HeartRateUtils getInstance() {
+        return HeartRateUtilsInstanceHolder.INSTANCE;
+    }
+
+    /**
+     * Singleton - to access this class use the static #getInstance()
+     */
+    private HeartRateUtils() {
+        updateCachedHeartRatePreferences();
+    }
+
+    public void updateCachedHeartRatePreferences(){
+        Prefs prefs = GBApplication.getPrefs();
+        maxHeartRateValue = prefs.getInt(GBPrefs.CHART_MAX_HEART_RATE, MAX_HEART_RATE_VALUE);
+        minHeartRateValue = prefs.getInt(GBPrefs.CHART_MIN_HEART_RATE, MIN_HEART_RATE_VALUE);
+    }
+
+    public int getMaxHeartRate(){
+        return maxHeartRateValue;
+    }
+
+    public int getMinHeartRate(){
+        return minHeartRateValue;
+    }
+
+    public boolean isValidHeartRateValue(int value) {
+        return value >= getMinHeartRate() && value <= getMaxHeartRate();
     }
 }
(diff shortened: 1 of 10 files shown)


Dikay900:
updated screenshots so we can better see the difference

cpfeiffer:
I completely agree that this is very unfortunate. I'm not sure that raising the minimum HR value in general for all devices is the correct solution here. Either we should make the value configurable or we should special case it for Mi Band 3.

Dikay900:
ok i will look into make it a configurable value, although i do not see where this value which is also used for checking whether we got a appropriate Heart Rate should not be raised to something like 30/40 in general.

ashimokawa:
Honestly I think it is ok to merge this as is. It wont change the raw database at all, we can still do something else later. And HR 10 is obviously not healthy ;)

cpfeiffer:
Thanks a lot. Unfortunately I still cannot merge this as is, because access to the preferences twice for every single sample would be a bit of a performance hit. The class is also used in GPXExporter, where hr is verified for every single trackpoint.

So we would need a cached version of the values and a preferences listener that updates them. Maybe we then need an instance of the HeartRateUtils class instead of just the static method.

Let me know if I should take this over or if you want to have a go at it.

Dikay900:
Thanks for the hint i did miss the usage in the GPXExporter class so this is indeed an issue. I will solve this later.

Dikay900:
I've updated my code.
We probably want the SharedPreferenceListener in a separate Service.
Do you have a hint where to put that listener. This is the easy but not very clean way currently i think.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
```diff
@@ -28,7 +32,42 @@
...
 
-    public static boolean isValidHeartRateValue(int value) {
-        return value > HeartRateUtils.MIN_HEART_RATE_VALUE && value < HeartRateUtils.MAX_HEART_RATE_VALUE;
+    private int maxHeartRateValue;
+    private int minHeartRateValue;
+
+    /**
+     * extra class used for synchronized access of the singleton
+     */
+    private static final class HeartRateUtilsInstanceHolder {
```
cpfeiffer: Can you please elaborate why we need the separate class?

If you put the static initialization directly into the `HeartRateUtils` class, this will happen at classloading time, which is perfectly synchronized. Could the static method be invoked before the static initializer is finished?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/export/GPXExporter.java
```diff
@@ -181,7 +181,7 @@ private void exportTrackpointExtensions(XmlSerializer ser, ActivityPoint point,
         long lowestDifference = 60 * 2 * 1000; // minimum distance is 2min
         for (ActivityPoint pointItem : trackPoints) {
             int hrItem = pointItem.getHeartRate();
-            if (HeartRateUtils.isValidHeartRateValue(hrItem)) {
+            if (HeartRateUtils.getInstance().isValidHeartRateValue(hrItem)) {
```
cpfeiffer: Same here, remember it outside the loop.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
```diff
@@ -28,7 +32,42 @@
...
+    public int getMaxHeartRate(){
+        return maxHeartRateValue;
+    }
+
+    public int getMinHeartRate(){
+        return minHeartRateValue;
+    }
+
+    public boolean isValidHeartRateValue(int value) {
+        return value > getMinHeartRate() && value < getMaxHeartRate();
```
cpfeiffer: Nitpicking, but shouldn't this be `>=` and `<=`?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
```diff
@@ -28,7 +32,42 @@
...
+     * Singleton - to access this class use the static #getInstance()
+     */
+    private HeartRateUtils() {
+        updateCachedHeartRatePreferences();
+    }
+
+    public void updateCachedHeartRatePreferences(){
+        Prefs prefs = GBApplication.getPrefs();
+        maxHeartRateValue = prefs.getInt("chart_max_heart_rate", MAX_HEART_RATE_VALUE);
+        minHeartRateValue = prefs.getInt("chart_min_heart_rate", MIN_HEART_RATE_VALUE);
```
cpfeiffer: Can you add those preference keys to `GBPrefs` and use them instead of scattering the keys? I know we have more of those, but I'd like to have fewer of them, not more :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractChartFragment.java
```diff
@@ -512,7 +510,7 @@ protected void refresh() {
                         }
                         activityEntries.add(createLineEntry(value, ts));
                 }
-                if (hr && sample.getKind() != ActivityKind.TYPE_NOT_WORN && HeartRateUtils.isValidHeartRateValue(sample.getHeartRate())) {
+                if (hr && sample.getKind() != ActivityKind.TYPE_NOT_WORN && HeartRateUtils.getInstance().isValidHeartRateValue(sample.getHeartRate())) {
```
cpfeiffer: Not sure if the getInstance() call would be optimized by the runtime or art. We could remember the instance outside the loop.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
Dikay900: This is used to be synchronized when calling HeartRateUtils.getInstance() the first time in two different threads (race condition).
This will lazy load the instance when the class is loaded.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/export/GPXExporter.java
Dikay900: yes this should be cleaner this way. Will rework this.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractChartFragment.java
Dikay900: yes this should be cleaner this way. Will rework this.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
Dikay900: yes you are right, just rewrote the exisiting function. Will change this as well.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
Dikay900: Ok did that.

Dikay900:
updated and squashed last commit

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/HeartRateUtils.java
cpfeiffer: AFAICS we don't need any synchronization here. If we would lazily create the instance inside the `getInstance()` method, then this would need to be synchronized. But we create the instance at classloading time already (in the static initializer), so no need for any further synchronization, nor for the inner class.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/DeviceCommunicationService.java
```diff
@@ -784,6 +785,9 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
                 mDeviceSupport.setAutoReconnect(autoReconnect);
             }
         }
+        if (key.equals("chart_max_heart_rate") || key.equals("chart_min_heart_rate")) {
```
cpfeiffer: Use the constant from `GBPrefs` here? :-)

Dikay900: yes you are correct <.< i shouldn't miss such things.
thanks

Dikay900:
Ok thanks for your comments, updated again.

cpfeiffer:
Woohoo :tada: :smile:

xandro0777:
Just a guess, with the optical sensor it seems next to impossible to measure any HR value under 30. To get a reasonable measurement I would think the sensor must read detect at least 4 beats and I have never seen the LEDs lightning for longer than a few seconds unless in continuous mode.

Issue #1156, opened 2018-07-13
[Device request] Lenovo Watch 9


KonradIT:
[Lenovo Watch 9](https://www.gearbest.com/smart-watches/pp_1835306.html?wid=1349303)

Supports HRM, notifications + calls, sleep tracking and alarms. Its pretty cheap as well.

[Android app](https://play.google.com/store/apps/details?id=com.pmpd.dmap.lenovo).

adambuczek:
+1
Watch looks great but the app is very poor.

rzetzsche:
+1 
Watch is pretty awesome but the app is a mess.

cpfeiffer:
Looks like @f3z is looking to add support for it. :+1: :+1: Just yell if you need help somewhere.

f3z:
Anyone know much about BLE pairing in terms of gadget bridge? I've managed to find the UUIDs needed for comms

cpfeiffer:
Have a look at https://github.com/Freeyourgadget/Gadgetbridge/wiki/New-Device-Tutorial where we describe discovery and pairing, among others.

joelbars:
@f3z I really want to help, I have a device too and I following the guide to add a new device, I didn't managed to find the UUIDs, is there anything that I can do to help? Maybe together we can finish this faster...

f3z:
@cpfeiffer I have read through the tutorial and have managed to get the app at least see the device but haven't had much luck with pairing. The watch requires the button to be pressed to complete the pairing process.

joelbars:
Hi @f3z, would be possible for you to share the UUIDs? Thanks.

f3z:
0000a800-0000-1000-8000-00805f9b34fb
0000a801-0000-1000-8000-00805f9b34fb
0000a802-0000-1000-8000-00805f9b34fb
0000a803-0000-1000-8000-00805f9b34fb

cpfeiffer:
@f3z But pressing the button does not do anything? I had added some bits to the tutorial regarding pairing yesterday (getBondingStyle()). What does your DeviceCoordinator#getBondingStyle() method return?

joelbars:
[...]

PR #1190, opened 2018-08-06
Watch 9: Initial support

maxirnilian:
Initial support for Watch 9 (hybrid smartwatch)

Working: 
- Notifications
- Alarms (maybe some further testing is needed to verify every alarm scenario)
- Setting step goal (progress can be checked on the watch face)
- Time calibration/sync

Some features are missing, but for now that is everything I found out.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -388,6 +388,12 @@
             android:name=".devices.pebble.PebblePairingActivity"
             android:label="@string/title_activity_pebble_pairing" />
         <activity
+            android:name=".devices.watch9.Watch9PairingActivity"
+            android:label="@string/title_activity_watch9_pairing"  />
+        <activity
+            android:name=".devices.watch9.Watch9CalibrationActivity"
+            android:label="@string/title_activity_watch9_calibration"  />
+        <activity
             android:name=".activities.charts.ChartsActivity"
             android:label="@string/title_activity_charts"
             android:parentActivityName=".activities.ControlCenterv2" />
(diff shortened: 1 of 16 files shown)


maxirnilian:
Hey, thanks for the quick review. Do I just have to commit the necessary changes to my branch to appear here for a re-review (this PR stuff is pretty new to me :see_no_evil:)?

The Watch is related to issue #1156 (Lenovo Watch 9), sorry missed that in the first comment. I didn't find more useful (or any official) device related stuff than those links in the issue. I would guess that there are more devices using the underlying hardware, but can't say that for sure. 

Sounds good, I can not add much more for now. What do you think would be the best practice? Maybe someone else owning this device would like to take a look on the missing features and/or improvements, so is it ok to have it on master or would it be better in a separate branch?

One additional question, GB seems to be "unable" to handle battery info correctly for devices having a battery state (OK or LOW) but not a battery level (always fires low battery -1% notification) should I open a separate issue for that or just change it so you can also review this here?

cpfeiffer:
I'm fine with merging this early. Now we even have an independent confirmation by @f3z that it works, so that's great!

After it's merged, just create new pull requests for further changes.

For the battery problem, please open a new issue. I don't quite understand if the problem is in BLE code or in the UI side or somewhere else, so let's discuss this in a separate issue.

cpfeiffer:
So I'm merging this. Thanks a bunch :tada:

maxirnilian:
Oh that was fast, I haven't even managed to apply your remarks yet. Maybe those can be merged at a later point, since the changes seem to be quiet small.

> So I'm merging this. Thanks a bunch :tada: 

No probs, I'm happy I could spend support to this project.

> For the battery problem, please open a new issue. 

OK, will do :+1:

cpfeiffer:
Right, the few changes can be quickly applied when you submit them.

PR #1196, opened 2018-08-08
Mi Band 3: New icon / Band screen unlock

joserebelo:
- Add a new icon for the Mi Bands 2 & 3.
- Add a preference to enable/disable the band screen unlock.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/huami/miband3/MiBand3Support.java
```diff
@@ -71,6 +75,45 @@ protected AmazfitBipSupport setDisplayItems(TransactionBuilder builder) {
         return this;
     }
 
+    @Override
+    public void onSendConfiguration(String config) {
+        TransactionBuilder builder;
+        try {
+            builder = performInitialized("Sending configuration for option: " + config);
```
joserebelo: I am just unsure about overriding onSendConfiguration, because performInitialized may be called twice in very quick succession. Can that be a problem?

cpfeiffer: In theory, that shouldn't be a problem :grimacing: 
Both calls will just add actions to the queue. The queue is properly synchronized. Initialization (if necessary) will only happen once.

joserebelo: Alright, that was my concern. Then everything should be okay.

ashimokawa:
@joserebelo 
Many thanks, tested, and works :+1:

Issue #1113, opened 2018-06-01
Mi Band 3 Support


chrisvella:
This is more for awareness at this stage but the Mi Band 3 has been announced. Info [here](https://www.androidcentral.com/xiaomi-mi-band-3-picks-upgraded-oled-panel-and-nfc-costs-just-30). It has many of the same features as the Mi Band 2 but also adds NFC.

danielegobbetti:
Let's hope it will use a protocol similar to miband2/amazfit/huami recent devices. Hopefully now that xiaomi opened real stores in Europe we will be able to get one sooner than in the past with other similar devices.

If you or anyone else happen to get one and want to help development, please chime in here. We can provide directions and hopefully (see my note above) getting the basics working in a short amount of time.

dan9550:
I'd be happy to buy one and test with it, in terms of development i'm not sure how useful i'd be. My Java game is not strong.

ashimokawa:
@dan9550 
Thanks! It would be really helpful if you could compile Gadgetbridge yourself, then we could have short test cycles.

dan9550:
Alright. I have a Mi Band 3 on the way, will update when it arrives.

IanDGibbons:
Me too. Should be here in a couple of weeks.  I'll setup a dev environment for GadgetBridge and get compiling.

ashimokawa:
@dan9550 
@IanDGibbons 
I just pushed initail Mi Band 3 support (blindly), just to separate from Mi Band 2 and enable firmware flashing.

chienkd:
@dan9550 Please support Miband 3 to flash firmware, res and font.

IanDGibbons:
The band has arrived and is now fully charged.  I have compiled a new Gadget Bridge with Mi Band 3 support and installed (old version backed up first!)

Device discovery did not pick it up at first but now it appears in the list.

The first 3 parts of the MAC address are: FA:35:ED

It will not connect.  The mi band 3 shows the same picture (attached) that it had before the pairing.

I downloaded the Mi Fit app and extracted all the firmware / font files .  Here is a list of filenames:

- HM0D.fw
- HM03.fw
- HM05.fw
- HM07.fw
- HM08.fw
- HuaShan.fw
- Mili.fw
- Mili_1a.fw
- Mili_beats.fw
- Mili_beats_p.fw
- Mili_chaohu.ft
- Mili_chaohu.fw
- Mili_chongqing.fw
[...]

PR #1202, opened 2018-08-10
Mi Band 3: Change display items

joserebelo:
Allows changing the Mi Band 3 display items.

The order currently depends on the iteration order of the `Set<String> pages`. The function already allows for the order to be set if a `List<String>` is provided with the proper order, for example, but I won't have the time to implement the UI for that any time soon, so this PR at least allows for the selection.

The 7th byte of COMMAND_CHANGE_SCREENS is always 0x06 (last position) probably reserved for some extra screen in the future, I tried to enable it without success.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/SettingsActivity.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/SettingsActivity.java
@@ -405,8 +405,6 @@ public class SettingsActivity extends AbstractSettingsActivity {
             }
         });
 
-
-        /*
         final Preference miBand3DisplayItems = findPreference("miband3_display_items");
         miBand3DisplayItems.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
             @Override
@@ -420,7 +418,6 @@ public class SettingsActivity extends AbstractSettingsActivity {
                 return true;
             }
         });
-        */
 
         final Preference corDisplayItems = findPreference("cor_display_items");
         corDisplayItems.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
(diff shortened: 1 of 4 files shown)


ashimokawa:
Thanks, will test today. My code that was there led to weird and unpredictable behaviour, let's try yours ;)

ashimokawa:
@joserebelo 
Other devices can also set the order, I tried with the Bip long ago, what your PR showed me is that unlike the Bip, the Mi Band 3 firmware seems to be picky about "holes" in the numbering of actually enabled items.

I will merge this but then revert the logic partly because of the non-deterministic order.

For proper reordering, my plan is to use the "AppManager" used by the Pebble to allow reordering of build-in system apps, it has a gui for that, but need to get rid of these Pebble-isms and probably switch it from the text file hack to a database table. We also already have an API for sending App order **void onAppReorder(UUID uuids[])**, this is because all Pebble apps  (even build in ones) have a uuid. We could reuse that an specify fake uuids for the Huami menu items. This would magically allow per-device settings.

And yes I am interested in what you captured. I had a really hard time with trial and error and I am still not sure if I do it correctly.

joserebelo:
@ashimokawa 
Just dawned on me that you may have notifications disabled for that one issue. I have uploaded the dump in #1113.

I had forgotten about the non-determinism caused by iterating the Set, sorry about that.

I was not aware of the AppManager, never used a Pebble with Gadgetbridge. Guess that will make it much easier and consistent across devices 👍

timonoj:
Hi...Just to confirm, Notifications and Weather are still not working on the Mi Band 3, is this correct?

Thanks!

By the way, I installed successfully the FW 46985, the RES 25278 and the fonts included in the Mi Fit apk v3.4.7.27. I believe it's not whitelisted yet.

joserebelo:
Both should be working already, this PR only allowed to show/hide the menu items.

ashimokawa:
All fixes have been released already, make sure to use 0.29.1

PR #1207, opened 2018-08-13
Support added for H3 watch strap

Dougal19:
Support added for H3 smart watch strap, which also uses JYou app to connect.


cpfeiffer:
Thanks a lot! :-)

cpfeiffer:
One thing: I hope that the device name prefix you added is not unique to your actual device and that the `0DE7` is actually a general prefix..

Dougal19:
Good point, unfortunately I only have one to test at the moment. I do have a friend who has ordered one and will check the prefix on theirs when they get it.

Dougal19:
There is also a H1 variant which I assume would also work with the right name added to the file. The photo here shows the name as H1-4D6D.
https://www.reddit.com/r/Watches/comments/8r3u3g/orient_bambino_smartwatch_best_of_both_worlds/

Maybe if the four digits at the end differ there is a way to support any combination? I'm not a programmer so not sure how that would work.

cpfeiffer:
Alright, we need to have a closer look at this, then.

PR #1212, opened 2018-08-15
Enable word wrap and set center horizontal alignment for legend into …

elagin:
…ActivitySleepChart and WeekSleepChart
My phone does not fit text charts_legend_heartrate in portrait mode.


cpfeiffer:
I'll merge this and see how this looks on other devices. Thanks!

PR #1213, opened 2018-08-15
Enable multiline title for CheckBoxPreference

elagin:


cpfeiffer:
This is because some strings are too long? Would it perhaps be easier to shorten the strings?

elagin:
For example, the Russian language consists of longer words than English and can not always be shortened. What for for each language to invent abbreviations, that precisely the text was located in the necessary place if it is possible simply to take enough place.

cpfeiffer:
Alright, thanks a lot, I merged this manually.

Issue #164, opened 2015-11-09
Display more sleep info


cpfeiffer:
I would like to see at what time I fell asleep, at what time I woke up.

nico202:
Also, in the default mi band app, you can change the time you went to bed/the time you woke up.

cpfeiffer:
Yes, good point.

cpfeiffer:
@elagin At least on my Bip the end time (= wakeup) is often very wrong. Can you check if this also happens for you? Ideally, the wakeup time should be the same as start time + duration of sleep.

elagin:
https://drive.google.com/file/d/1MmfSVwHQJnubGDZqcy1l_r2s5Aw-a5gY/view?usp=drivesdk

cpfeiffer:
Yeah, 23:53 + 7:45 != 8:03

What happens if you toggle through the days with the arrow buttons at the bottom? For me the wake up time is sometimes a minute before the fall-asleep time.

elagin:
It seems I found one of the problems. In the data there is information about a sleep lasting 1-2 minutes, and at that time I was clearly not sleeping:
time | raw_kind | text
01:37 | 12 | Activity
01:38 | 12 | Activity
01:39 | 12 | Activity
01:40 | 12 | Activity
01:41 | 11 | Deep sleep
01:42 | 12 | Activity
01:43 | 12 | Activity
01:44 | 11 | Deep sleep
01:45 | 11 | Deep sleep
01:46 | 12 | Activity
01:47 | 12 | Activity
It is necessary to determine the minimum duration of sleep in order to exclude false information.

elagin:
Also, there can be several dreams in one night. For example, that night, that on my screenshot at 23:50 began recording a deep / fast sleep. From 06:05 the recording of activity began. From 06:39 the dream began again, which lasted until 08:03. Therefore, the total number of minutes of sleep was less than the time elapsed from 23:50 to 08:03. The current algorithm does not take into account the possibility of breaks between dreams.

pixelipo:
Some padding is missing in front of that string?

PR #1216, opened 2018-08-16
Display start and end sleep.

elagin:


elagin:
For issure  #164

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/ActivityAnalysis.java
```diff
@@ -95,6 +94,9 @@ ActivityAmounts calculateActivityAmounts(List<? extends ActivitySample> samples)
                 }
             }
 
+            amount.setStartDate(sample.getTimestamp());
+            amount.setEndDate(sample.getTimestamp());
```
cpfeiffer: It's a bit weird to set start and end to the same value, but I get the point.

PR #1220, opened 2018-08-17
Mi Band 3: Night Mode

joserebelo:
Tested on firmware 1.4.0.6, probably not compatible with previous firmwares.


cpfeiffer:
Wow, very nice! I'll pass this to @ashimokawa for testing, but code-wise it's :+1:

ashimokawa:
@joserebelo 
Thanks, and sorry for the delay. Will test tomorrow!

ashimokawa:
@joserebelo 
Tested, and works

A few remarks:
- Since we do not send sunrist and sunset time yet, it does not make sense to set this option, I will disable it for now. We do use a library to calculate sunrise and sunset offline, that would make sense to use there. I want to wire that up as soon as I have time
- I do not like the MiBandConst class, it mixes up Mi Band 1 and 2 which have almost nothing in common, also I do not like to name things "MI2" oder "MI3" is ist calles "Mi Band" and not Mi ;), this is not your fault you followed the code faithfully, unfortunately our code is is mess. I started to rename stuff and also made "HuamiSupport" where we should implement everything that is used for more than one device (Currently as a hack Mi Band 3 and Cor are derrived from AmazfitBipSupport because the Bip was the first modern Huami device which added a lot of new stuff, we did not know that a horde of new (semi-)compatible devices will pop up). Long story, I will merge anyway, but we should keep in mind that this is not how it should be ;)

PR #1222, opened 2018-08-18
Support minimum time between notifications

joserebelo:
Simple implementation that allows the user to specify a minimum time between notifications, preventing spam from chat apps.

Possible future improvements: 
- Per-app configuration?
- Reset on screen on? (optionally)


cpfeiffer:
We actually have some kind of throttling already, see ServiceDeviceSupport.java. Having this configurable is obviously much better than the hardcoded limit.

joserebelo:
Oops, I looked around and yet totally missed that, although this works slightly differently, throttling each application as well.

PR #1225, opened 2018-08-20
Add share log file. Issue #1146

elagin:
By issue #1146 Tested on gmail, evernote, google drive.


cpfeiffer:
Thanks, that's pretty cool! :+1: 

We might want to add some kind of warning that the log could contain sensitive information. Any thoughts? @ashimokawa @danielegobbetti @IzzySoft and everyone else?

MagicFab:
Muic information, MAC address come to mind. I'd expect this to be removed for me or have an option to do so when submitting logs, but not when saving them for myself.

IzzySoft:
A hint should be given indeed. Maybe a dialog stating that, giving the options to cancel the action, to do it this once, or always accept (i.e. remember the decision for the next time).

I'd like the idea @MagicFab brought up, to (optionally) have sensitive information removed. But a) what belongs to that category, and b) how can we automatically identify it? The MACs of used devices is easy as t's in the configuration. But what else could there be? If we automatically prune the logs, users might expect a "perfect prune" which is hard to satisfy. So if we try that, it should be made clear there might be data left and the user has to check again manually to make sure.

cpfeiffer:
Yes, I'm afraid we can't automate a perfect pruning of sensitive information. The MAC itself is a start, but may not be sufficient. Imagine a log containing discovery information, where vibratissimo-like devices are captured, for example :smile:

So, we could "anonymize" all MACs by default and offer the user a way to preview and save the log file.

Now I'm not sure if @elagin feels like doing all this... :smile_cat:

elagin:
I'm not really scared if I share my log file somewhere and it will be known that at 10:35 my heartbeat was 78bpm. Yes, this file can contain confidential information, but just someone in the subway can look at how much I woke up this morning. What to do with it? In some places, high-resolution surveillance cameras operate that can see the input of the password.
In the suggestion when I press the "share logfile" button to display confirmation "the log file can contain confidential information, do you really want to share it."

danielegobbetti:
I agree that a warning should be there and there is realistically no pruning that could be performed.

MagicFab:
@elagin please avoid arguments along the ["I have nothing to hide" argument](https://en.wikipedia.org/wiki/Nothing_to_hide_argument) logic, by that logic the project is irrelevant. I want my privacy in all aspects of this project, including logs I may have to submit to help in debugging issues.

A good warning would indicate the logs contain **lots** of personal information and suggest to anonymize it manually before sharing. I've added that to [the FAQ entry on log files](https://github.com/Freeyourgadget/Gadgetbridge/wiki/Log-Files), but an in-app warning or link to the FAQ would be helpful too.

The FAQ entry on logs could also indicate how to send these files privately to a developer, upon request. Once it's decided how to do so I can help document this.

elagin:
Then I suggest before the share button place the TextView with the text:
> Please keep in mind Gadgetbridge log files may contain lots of personal information, including but not limited to health data, unique identifiers (such as a device MAC address), music preferences, etc. Consider editing the file and removing this information before sending the file to a public issue report.

elagin:
@IzzySoft  I add dialog warning for share log-file.

IzzySoft:
Thanks, @elagin – but I just wonder whether all commits here are related to this PR? Shouldn't that rather be multiple separate PRs? Unrelated to the log share, I see:

* updates to documentation
* Night Mode for Mi:Band 3
* translations
* Support minimum time between notifications

So IMHO, this should be 5 separate PRs. But not being a maintainer here, I leave that decision to @danielegobbetti, @cpfeiffer and @ashimokawa to decide.

elagin:
I apologize, I'll do it, mergers from the master were added. Let's do a pure PR.

PR #1226, opened 2018-08-20
I suggest displaying sleep and not worn entries as constant values.

elagin:
gadgetbridge

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractChartFragment.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/activities/charts/AbstractChartFragment.java
@@ -93,6 +93,8 @@ import static nodomain.freeyourgadget.gadgetbridge.activities.HeartRateUtils.isV
 public abstract class AbstractChartFragment extends AbstractGBFragment {
     protected final int ANIM_TIME = 250;
 
+    public static final float Y_VALUE_NOT_WORN = 0.1f;
+
     private static final Logger LOG = LoggerFactory.getLogger(AbstractChartFragment.class);
 
     private final Set<String> mIntentFilterActions;
@@ -474,7 +476,7 @@ public abstract class AbstractChartFragment extends AbstractGBFragment {
                             notWornEntries.add(createLineEntry(0, ts));
                             activityEntries.add(createLineEntry(0, ts));
                         }
-                        deepSleepEntries.add(createLineEntry(value + SleepUtils.Y_VALUE_DEEP_SLEEP, ts));
+                        deepSleepEntries.add(createLineEntry(SleepUtils.Y_VALUE_DEEP_SLEEP, ts));
                         break;
                     case ActivityKind.TYPE_LIGHT_SLEEP:
                         if (last_type != type) {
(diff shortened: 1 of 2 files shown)


Dikay900:
I would rather like to see a full 1f as the constant so we got a full bar for these activities.
What do you think?

elagin:
It seems to me that this does not look beautiful.
2018-08-27-16-44-27-340 gadgetbridge

elagin:
The presence of "Not worn" in this graph is another matter. Green activity is also not needed here. in a dream it should not be. Specifically in this PR, I propose to establish a constant height for sleep.
While I'm waiting for the merge of this PR, I am working to improve this screen. I hope you like it:
image

LuccoJ:
Well... I don't. For a start, it's well-known that deep sleep is not recognized correctly at least for Huami devices, so aside from the heart rate, you'd be removing the *only* information about the amount of activity during sleep. I have absolutely no idea why one would *not* want to see how much they were moving during dreams as well as during deep sleep. That's an important part of sleep monitoring.

I personally hope this PR is rejected. If it is accepted, I will have no choice but to move to a different app, if I want to see the data I want to see.

elagin:
@LuccoJ If you need to see activity during sleep, you can leave it. (and you can turn on / off the settings). The pie chart still does not take it in calculating the duration of sleep, it considers only records of deep / light sleep. Activity outside sleep you also need to see?

Dikay900:
I see and have to agree to @LuccoJ. Yes we should still be able when and how much we move in sleep since the sleep recognition is currently not postprocessed by GB. Considering this i would also reject this.
This could be adressed when adding an additional logic behind the sleep reporting (Light sleep / Deep Sleep / REM).

elagin:
Something we went to the side. In this PR it is proposed to make the graphics of light and deep sleep of the same height.

Dikay900:
well since the height is used for the activity which we currently still need for further sleep investigating we still need the value of the samples but i agree the base height could be adjusted for both to be the same when there is no activity. Same goes for the not worn entries.

elagin:
@Dikay900, thank you for your support, we are waiting for the PR to be accepted.

cpfeiffer:
I think most of us want to see the different (non-constant) values for both light and deep sleep, so we cannot accept the PR as is. I'm very much open for improvements, but hiding away the different values is not an option.

An idea: maybe it would look OK, if we would colourize all activity types as full bar (value 1). Then we could plot a light line as the activity value over those colour bars.

WDYT?

LuccoJ:
@cpfeiffer *personally* I would be okay with that, using very "transparent" colors for the background... but I think this should still be toggleable, because it may have a dramatic impact on users with reduced visual acuity.

elagin:
@cpfeiffer 
I do not understand how you suggest this should look. Please draw.
Please explain what data is contained in the value sleep? Why does the bracelet send them not in the form of a bool, because I either sleep or not. After all, he decided that at this point in time I sleep, for example, a deep sleep.
Or are these different values ​​of sleep the same as "activity" (which is drawn in green), when I do not sleep?

LuccoJ:
@cpfeiffer Oh wait, I had understood something different. It's subjective, but I do not really like your idea now that I understand it: there would end up being two lines, one for heartrate and one for activity, and while the information is the same after all, I think it would make matters more confusing.

Instead what I had thought you were proposing (and in my mind it looks better) is something like your first picture, but without the actual line, and instead, the sections of colors above the (no longer existing) line would be less bright / less saturated / less something than the sections below.

I'll re-use your picture to create the mockup below, please don't DMCA yourself due to my infringement of your picture's copyright, though!

lines

I didn't manage to remove the line completely, but imagine it as either not being there, or being only there very subtly (thin and desaturated).

elagin:
@cpfeiffer Thanks for the visualization, but in reality the values ​​of the activity vary greatly in every following record. Therefore, do not get such a smooth envelope line.

PR #1228, opened 2018-08-20
update README fdroid badge

TacoTheDank:


danielegobbetti:
Why the change? Why hotlinking the image? Is this desired by f-droid? Do you have links to prove it?

IzzySoft:
While changing the URL from `https://f-droid.org/repository/browse/?fdid=nodomain.freeyourgadget.gadgetbridge` to `https://f-droid.org/packages/nodomain.freeyourgadget.gadgetbridge/` is certainly correct (the former is just redirecting to the latter for compatibility, due to website changes which happened a while ago), this shouldn't be done as "plain HTML", but Markdown should be kept.

Further I agree with @danielegobbetti and see no reason to deep-link the logo.

So what's your idea behind that, Taco? It's always a good idea to include some comment to let people know the intention of the proposed change :wink:

TacoTheDank:
just replacing the old "get it on f-droid" with the new one. its not at all required or anything

TacoTheDank:
oh god ignore all of those

IzzySoft:
You still use `` instead of `img`. If I translate that back to Markdown, it comes up to what it originally was. I've got no clue what the change should be good for.

TacoTheDank:
because if i just use the other one, the f-droid badge would be super huge. i have no idea how to specify a height in markdown

IzzySoft:
Ah, missed that. And right, in Github's Markdown one cannot specify image sizes. But one could use resized images :wink:

danielegobbetti:
I asked [upstream](https://gitlab.com/fdroid/artwork/issues/16) more info re the hotlinking.

danielegobbetti:
@TacoTheDank I'm sorry but it was faster to do directly the changes according to upstream than it would have been to cherry pick among your commits. I'm looking forward to further contributions from you though! Thanks

Issue #976, opened 2018-02-05
Suggestion: RTL support


doryashar:
Hi, i use amazfit bip and its really annoying that it does not support rtl.
Since we can modify font files now, my suggestion is to do the ltr-rtl translation of the notification inside the app (for the required languages)

Thanks.

NotAFile:
It sounds like you might be referencing me documenting the HMZK (mi band 2) font files.

Unfortunately, as far as I know, the HMRES files used by the mi band 2 are not jet documented. I just had a look at them and they look far, far more complicated than the mi band 2 files. So you might need to wait a bit for someone to reverse those. I might give it a shot, but no promises.

doryashar:
you did not understand me
i have installed hebrew font on AMAZFIT BIP
but, unlike most of the languages which are written Left-To-Right, hebrew is written Right-to-Left
so all hebrew notification are shown backwards,
Therefor, my suggestion is fixing the notification module to send hebrew characters from right to left.

NotAFile:
oh, okay, I understand what you mean. But even if you reverse the characters, it will still flow LTR. Also, doing that won't work with languages like arabic. But I understand what you mean now.

gruyon:
Hi all.
Can someone help me adding the font to my bip or will it still be backward ?
Should I stay with my basic eng version.

roigreenberg:
Hi.
I wish to work on this, and my idea is to "convert" the notification before it being sent to the phone.

Can you please help me and tell me where in the code should I look?
@ashimokawa

cpfeiffer:
Have a look at the LanguageUtils.java class, especially the methods about transliteration. At some point we're going to make this a bit more configurable and structured, but this is probably one of the better places to look at implementing RTL support.

roigreenberg:
Thanks.
Actually I found it yesterday, and done it on both languageType and
NotificationType(inside truncate)

I managed to reverse the message, the only issue left is that it now starts
from bottom to top.
I need a way to know(or estimate) the width of the line..

בתאריך יום ב׳, 20 באוג׳ 2018, 23:22, מאת Carsten Pfeiffer ‏<
notifications@github.com>:

roigreenberg:
I'm working on amazfit bip (and mi band 3).

I thought to simply take an average of ~15 characters per line.
In both cases I will need force it to write new lines with \n (hopefully it
supported..)



[...]

PR #1233, opened 2018-08-23
Adding RTL support

roigreenberg:
Regarding: Suggestion: RTL support #976

I added RTL support.
What it does is at the same place as "transliteration" happen, it takes the message, and flip the order of the characters so at the band/watch it will be shown correctly.

In order to also make it from top to bottom, I split the text into lines separated by a new-line symbol.
The width of the line can be changed from the settings.

The next step is to be able to connect Arabic font (convert it to contextual  form)


roigreenberg:
Added support in Arabic contextual form

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
```diff
@@ -0,0 +1,466 @@
...
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @return true if the char is in the rtl range, otherwise false
+     */
+    static Boolean isArabic(char c){
+        for (Pair<Character, Character> rang: arabicRange) {
```
cpfeiffer: Similary, I wanted to propose directly using `char` instead of `Character` to avoid all the boxing and unboxing all the time, but from shortly glancing at the code, I cannot tell which one would be better. Can you have a look?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
```diff
@@ -0,0 +1,466 @@
...
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @return true if the char is in the rtl range, otherwise false
+     */
+    static Boolean isRtl(char c){
+        for (Pair<Character, Character> rang: rtlRange) {
```
cpfeiffer: Could we use `Character.getDirectionality()` instead?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: Pair can't get char.
Also, there are places I wanted the ability to send null, and that also require Character.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: I will try.
From what I see, I'm not sure it's cover *every* Hebrew and Arabic characters, but I'm not sure if those characters are really important.
I will do some testing.

roigreenberg:
Can you instruct me how to do the tests?
I will now prepare some sentences.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
cpfeiffer: Yes, you would simply use a `char[2]` instead of `Pair<Character,Character>`. If you make use of `null` values, you have to be careful, since unboxing a `null` `Character` will lead to a `NullPointerException`.

In any case, please try to use *either* `char` *or* `Character` as far as possible, i.e. avoid autoboxing back and force for performance reasons.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
cpfeiffer: Thanks! You certainly know more about this topic than me. I only learned about this method yesterday while reading up on this.

cpfeiffer:
re 1) having this configurable should be fine for now, so one could set a different value for a different device.

re 2) Have a look at class `LanguageUtilsTest`, it's pretty easy. You could test `RtlUtils` in the same way, although you would have to make parts of it `public`, then.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: I will review it today to see all the cases.
If I don't really need null, then I will change it to char. Otherwise, Character.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: I investigate this last night, and it seems I can use it.
What I changed is the implementation of isRtl/isHebrew and such to use this function instead of searching.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: I need Character for maps, so I switch every char to Character

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/RtlUtils.java
roigreenberg: Done

roigreenberg:
Done everything, only test cases are missing.

cpfeiffer:
:+1: Please yell if you have some testcases, if if you need any help with them.

roigreenberg:
Added tests.
I think I cover most of the cases..

What the next steps?

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/test/RtlUtilsTest.java
```diff
@@ -0,0 +1,368 @@
...
+        result = RtlUtils.isArabic(english1);
+        assertFalse("Is Arabic Character failed", result);
+    }
+
+    @Test
+    public void testIsLtr() throws Exception {
+        boolean result = RtlUtils.isLtr(english1);
+        assertTrue("Is Ltr Character failed", result);
+
+        result = RtlUtils.isLtr(english1);
```
cpfeiffer: This should be english2, I guess.

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/test/RtlUtilsTest.java
```diff
@@ -0,0 +1,368 @@
...
+        result = RtlUtils.isRtl(punctuation1);
+        assertFalse("Is Rtl Character failed", result);
+    }
+
+    @Test
+    public void testIsPunctuation() throws Exception {
+        boolean result = RtlUtils.isPunctuations(punctuation1);
+        assertTrue("Is Punctuation Character failed", result);
+
+        result = RtlUtils.isPunctuations(punctuation1);
```
cpfeiffer: And this probably punctuation2.

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/test/RtlUtilsTest.java
roigreenberg: Fixed. 
Thanks.

app/src/test/java/nodomain/freeyourgadget/gadgetbridge/test/RtlUtilsTest.java
roigreenberg: Fixed. 
Thanks.

roigreenberg:
Hi @cpfeiffer .
Is there any ETA for the next release?
There are many who waits for this update..

cpfeiffer:
Hopefully tonight. Sorry it took so long.

roigreenberg:
Great!

בתאריך יום א׳, 16 בספט׳ 2018, 9:40, מאת Carsten Pfeiffer ‏<
notifications@github.com>:

cpfeiffer:
I just found an ugly issue that I want to fix, so the release will be delayed until tomorrow (trying hard to fix it by then). Sorry.

roigreenberg:
That totally fine. Thanks for the update!

roigreenberg:
Hi @cpfeiffer . I saw you create new release a few days ago, but I don't see it in F-Droid. is it in some tests or something?

ashimokawa:
It is on fdroid now.
For questions about the fdroid release process check the wiki. This question was already asked countless times.

roigreenberg:
Thanks! Now I know for next time.

PR #1246, opened 2018-08-29
Improved Persian transliteration

haditim:
Persian (Farsi) transliteration was incomplete and most of the characters were proposed according to Arabic transliteration which would be confusing for Persians. I changed some of them and added new ones like question mark, etc.. Please note that Persian has several "z" and "s" since the language alphabet share a lot of similarity with Arabic. However, these are all pronounced the same way in Persian and it makes more sense to have the same character for all in each group.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/LanguageUtils.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/LanguageUtils.java
@@ -76,9 +76,13 @@ public class LanguageUtils {
             put('و', "w"); put('ي', "y"); put('ى', "a"); put('ﺓ', "");
             put('آ', "2"); put('ئ', "2"); put('إ', "2"); put('ؤ', "2"); put('أ', "2"); put('ء', "2");
 
-            // Farsi
-            put('پ', "p"); put('چ', "ch"); put('ڜ', "ch"); put('ڤ', "v"); put('ڥ', "v");
-            put('ڨ', "g"); put('گ', "g"); put('ݣ', "g");
+            // Persian(Farsi)
+            put('ا', "a"); put('ب', "b"); put('پ', "p"); put('ت', "t"); put('ث', "th"); put('ج', "j"); put('چ', "ch"); 
+            put('ح', "h"); put('خ', "kh"); put('د', "d"); put('ذ', "z"); put('ر', "r"); put('ز', "z"); put('ژ', "zh"); 
+            put('س', "s"); put('ش', "sh"); put('ص', "s"); put('ض', "z"); put('ط', "t"); put('ظ', "z'"); put('ع', "e"); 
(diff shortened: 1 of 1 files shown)


cpfeiffer:
Thanks! Can you please update the test case for that? It fails with your changes:
> nodomain.freeyourgadget.gadgetbridge.test.LanguageUtilsTest > testStringTransliterateArabic FAILED
    org.junit.ComparisonFailure at LanguageUtilsTest.java:42

haditim:
Thanks for replying.
Arabic and Persian share many alphabets. Unfortunately, setting the alphabets for Persian in  
`app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/LanguageUtils.java`
 leads to an error in function `testStringTransliterateArabic()` assertion in 
`app/src/test/java/nodomain/freeyourgadget/gadgetbridge/test/LanguageUtilsTest.java`
since it changes the output string which wouldn't match the expected one anymore.
Is there any way to check the language of the notification string ahead of transliteration to use Arabic or Persian transliteration function accordingly? If not this would lead to a conflict when languages share characters.

haditim:
Alternatively, the Arabic transliteration could be modified. To me transliterating alphabets to digits does not make sense. Instead, the phonetics which mimic the real sound should be used as I did for Persian.

cpfeiffer:
Maybe @tstein has some input on this?

tstein:
Thanks for the heads-up. Just so you guys know where I'm coming from - I'm not an Arabic speaker, but I have Arabic-speaking family and a couple of non-Arabic-speaking smartwatches. :) I added some letters for Farsi based on a naive reading of wikipedia, in the hope that it would be useful, and I apologize for anything I got wrong.

Unfortunately, neither the UI for transliteration nor the code that does it are complex enough to distinguish languages. It's just replacing (non-ASCII) characters with (ASCII) strings. It's not possible to have different transliterations for the same character or to selectively enable mappings for a particular language without some new code. For the moment, I think we need to agree on a single set of mappings for both languages. Fortunately, most of the ones you proposed are the same. I'll start a review and leave comments on the ones that differ.

re: numerals, I did this as a strict implementation of the [Arabic chat alphabet](https://en.wikipedia.org/wiki/Arabic_chat_alphabet). Its usefulness varies by letter: I'm happy to change غ to 'gh', but want to keep ع as '3'. Please see my line comments and maybe we can make everyone reasonably happy. :)

tstein:
Whoops, I just noticed #1250 after hitting "comment". I'll see you folks over there.

haditim:
Thanks for the reply. Could it be possible to add another option next to transliteration in settings for switching between Persian and Arabic?
I actually kept this pull request open for a non-temporary solution for this problem as opposed to #1250.

tstein:
@cpfeiffer, I think that we wanted to leave this one in limbo for the moment. Would you mind reverting this merge?

cpfeiffer:
Indeed, but see my last comment to #1250.

tstein:
Ah, I'm sorry. I thought it was an explicit merge. My bad. :)

cpfeiffer:
Never mind. I was also confused at first :-)

PR #1249, opened 2018-09-01
Roidmi: fix duplicated DeviceType key

joserebelo:
Fix duplicated DeviceType key introduced by the #1179 rebase.


cpfeiffer:
Thanks :-)

PR #1253, opened 2018-09-05
Fix caller ID for incoming calls on Android 9

### LINKED ISSUE #1252: Incoming calls show "Unknown" caller on Android 9
- [X] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [X] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [X] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:

After the upgrade to Android 9.0 on my Pixel device, incoming calls do no longer show the caller on the wearable. Instead the caller is identified as "Unknown".

This is most likely due to a behavior change in Android 9.0 concerning privacy. See the section "Restricted access to phone numbers" right here:
https://developer.android.com/about/versions/pie/android-9.0-changes-all#privacy-changes-all

>**Restricted access to phone numbers**
>
>Apps running on Android 9 cannot read phone numbers or phone state without first acquiring the READ_CALL_LOG permission, in addition to the other permissions that your app's use cases require.
>
>Phone numbers associated with incoming and outgoing calls are visible in the phone state broadcast, such as for incoming and outgoing calls and are accessible from the PhoneStateListener class. Without the READ_CALL_LOG permission, however, the phone number field that's provided in PHONE_STATE_CHANGED broadcasts and through PhoneStateListener is empty.
>
>To read phone numbers from phone state, update your app to request the necessary permissions based on your use case:
> - To read numbers from the PHONE_STATE intent action, you need both the READ_CALL_LOG permission and the READ_PHONE_STATE permission.
> - To read numbers from onCallStateChanged(), you need the READ_CALL_LOG permission only. You don't need the READ_PHONE_STATE permission.

#### Your wearable device is:

Pebble 2 HR (silk, FW 4.3)

#### Your android version is:

Android 9.0

0.29.1

skweee:
Fixes #1252

This fixes the caller id detection for Android 9. With the added `READ_CALL_LOG` permission, `PhoneCallReceiver` actually receives two broadcasts for each incoming call. The new guard in `PhoneCallReceiver` makes sure that only the broadcast which actually contains a populated `TelephonyManager.EXTRA_INCOMING_NUMBER` is processed. The other broadcast can safely be ignored according to the `TelephonyManager` documentation:
https://developer.android.com/reference/android/telephony/TelephonyManager#ACTION_PHONE_STATE_CHANGED

I could only test this change on Android 9.0 (Google Pixel), where it fixes #1252. Behavior on older Android versions needs to be tested before the merge.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
     -->
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
+    <uses-permission android:name="android.permission.READ_CALL_LOG" />
     <uses-permission android:name="android.permission.READ_PHONE_STATE" />
     <uses-permission android:name="android.permission.PROCESS_OUTGOING_CALLS" />
     <uses-permission android:name="android.permission.CALL_PHONE" />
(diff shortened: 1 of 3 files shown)


cpfeiffer:
This certainly looks good to me, so I'm merging this in, early so we can all test this.

danielegobbetti:
Will test on 8.1 soon (TM)

PR #1256, opened 2018-09-06
Calculation and display of weekly balance by steps and time of sleep.

elagin:


cpfeiffer:
Sounds good to me, but the balance calculation looks a little brittle and hard to understand to me. I'd suggest just declaring a `getBalance(ActivityAmounts activityAmounts)` method in the abstract superclass and calculating the value in the respective subclasses. No need for a member `mBalance` at all.

cpfeiffer:
Or, in case you don't have access to `ActivityAmounts` for every `getBalance()` call, calculate the value once when you do have access and just return the value later.

elagin:
@cpfeiffer 
I considered your recommendations. It is impossible to get rid of mBalance, it needs to be accumulated in Pie and every day. Maybe you should now take this PR, and then think about improving this code?

cpfeiffer:
Here's an explanation of my changes:
- a getter method is not typically supposed to change any internal state (basically like const-method in C++)
- the *Fragment classes are pretty much stateless, the data that they show is calculated in a background thread and stored in the `ChartsData` instance
- the calculation, accumulation and storage of the balance value should not happen in an unrelated  method (which also does not change any state)
- better accumulate the state (balance) on the outside (the caller) instead of in the method itself

One possible improvement that is left: if there is data is missing in the week for one or more days, the balance value is not very helpful (e.g. telling me at the begining of the week that I lack >40h of sleep).

PR #1257, opened 2018-09-06
Group language settings.

elagin:


cpfeiffer:
Makes sense, thanks a lot :+1:

PR #1292, opened 2018-10-02
ZeTime Configuration Support

Lightwars:
Now I would like to enable some options to configurate the ZeTime:
- do-not-disturb scheduling
- setting alarms
- automatic heart-rate measurement and heart-rate limits
- inactivity alert
- wearside of the watch
- wrist flick to en-/disable the display
- daily goals/targets
- date/time format, analog mode, calories type, activity tracking
- select signal types for different notifications

The options aren't complete yet and there is some room for improvement, but I would like to have you to have a look at this for now.

Please state if the daily goals in "about me" should be seperated for the ZeTime only?
Is it possible to wait for a submenu to be leaved to activate the preference change listener and send the whole config to the watch?


cpfeiffer:
@Lightwars Shall we merge this or do you want to change some things before?

Lightwars:
@cpfeiffer I want to change some things before I think this is ready. Some preferences should be reordered or maybe some more subscreens for them.
I wanted to read the actual preferences from the watch, but I couldn't figure out how to set them, although I used the SharedPreferences.Editor... If I have time I will update the code.

Lightwars:
@cpfeiffer I have been working on this lately, I feel it is ready to be merged.
Beware of this [a755318](https://github.com/Freeyourgadget/Gadgetbridge/pull/1292/commits/a7553188f92669f8e5785fc6759936cad5998bc7) commit, which I have done, to reading the preferences off the watch, when opening the ZeTime menu. Is it okay?

cpfeiffer:
Thanks so much! I will review tomorrow.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimePreferenceActivity.java
```diff
@@ -0,0 +1,461 @@
...
+                GBApplication.deviceService().onSendConfiguration(ZeTimeConstants.PREF_DATE_FORMAT);
+                return true;
+            }
+        });
+
+        final Preference inactivityAlert = findPreference(ZeTimeConstants.PREF_INACTIVITY_ENABLE);
+        inactivityAlert.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
+            @Override
+            public boolean onPreferenceChange(Preference preference, Object newVal) {
+                invokeLater(new Runnable() {
```
cpfeiffer: Can you add a short comment in the code why some listeners use invokeLater() and some don't?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimePreferenceActivity.java
```diff
@@ -0,0 +1,461 @@
...
+                    @Override
+                    public void run() {
+                    GBApplication.deviceService().onSendConfiguration(ZeTimeConstants.PREF_INACTIVITY_KEY);
+                    }
```
cpfeiffer: And since many of them send the same PREF_INACTIVITY_KEY, why not reuse one listener instance?

And for the other cases, one could probably use a method like this
```
private void addPreferenceHandlerFor(final String preferenceKey) {
	Preference pref = findPreference(preferenceKey);
	pref.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
		@Override
		public boolean onPreferenceChange(Preference preference, Object newVal) {
			invokeLater(new Runnable() {
				@Override
				public void run() {
					GBApplication.deviceService().onSendConfiguration(preferenceKey);
				}
			});
			return true;
		}
	});
}
```

and then just add the calls:

```
addPreferenceHandler(ZeTimeConstants.PREF_TIME_FORMAT);
addPreferenceHandler(ZeTimeConstants.PREF_ANTI_LOSS_SIGNALING);
addPreferenceHandler(ZeTimeConstants.PREF_SMS_SIGNALING);
```
etc. Saves probably 80% of the lines in the class.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/ActivityUser.java
```diff
@@ -127,11 +136,38 @@ private void fetchPreferences() {
...
+
+    public int getCaloriesBurnt()
+    {
+        if (activityUserCaloriesBurnt < 0) {
+            activityUserCaloriesBurnt = defaultUserCaloriesBurnt;
+        }
+        return activityUserCaloriesBurnt;
+    }
+
+    public int getDistanceKMeters()
```
cpfeiffer: I already see people asking for fractional values of km... Better use just meters here.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
```diff
@@ -70,7 +76,7 @@
     private final GBDeviceEventBatteryInfo batteryCmd = new GBDeviceEventBatteryInfo();
     private final GBDeviceEventVersionInfo versionCmd = new GBDeviceEventVersionInfo();
     private final GBDeviceEventMusicControl musicCmd = new GBDeviceEventMusicControl();
-    private final int sixHourOffset = 21600;
+    private final int sevenHourOffset = 25200;
```
cpfeiffer: Might your change from sixHourOffset to sevenHourOffset have something to do with dayllight savings?

cpfeiffer:
> Please state if the daily goals in "about me" should be seperated for the ZeTime only?
No, I'm pretty sure they will be useful for other devices as well. At some point we will need to mske it more clear, which preferences apply to which device and which are mandatory (for which device).

> Is it possible to wait for a submenu to be leaved to activate the preference change listener and send the whole config to the watch?
Yes, I'm pretty sure we already do something like that for a different device. Can't find it at the moment, though..

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
Lightwars: I think the firmware of the watch was changed and gives us timestamps which are UTC -7h, now.  My timezone is UTC+1h. The watch shows correct time, but the timestamps reported are not right.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/model/ActivityUser.java
Lightwars: Okay, I could change it to meters. But the ZeTime watch goals accepts only kilometers.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimePreferenceActivity.java
Lightwars: Sure, I think I get your idea. I will try it out, saving some lines is always a good choice. ;)
This my first project in android app and java development, so I'm still learning by looking at the code of other devices here.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimePreferenceActivity.java
Lightwars: Sure, I will think about it, if it is needed and why or I will delete those lines.

Lightwars:
> 
> 
> Wow, impressive, great job! muscle
> I only have a few remarks (see below).
> 
> I'm not 100% sure about `onReadConfiguration()`. AFAIU your preferences activity asks for the values to be read, and a little later (async) they are written into the preferences. Does the activity then update its view to reflect the new values? And I hope that the preference-change-listeners will not fire to send the values back to the device?
> 
Yes, the values are written async into the preferences and at the moment the activity is not updated to reflect the new values. Have you good a idea how to tell the activity to update? I have forgotten to look at this and how to do it...
The prefernce-change-listeners are not fired, but I will check it ASAP.

> Are all those preferences configurable on the ZeTime watch? And you apply a device-first approach (i.e. the device configuration would override what's saved in Gadgetbridge's preferences)?
> 
> Thanks a lot for your work! heart

Not every preference is configurable on the ZeTime watch, but most of it. Yes, I have done a device-first approach, cause the user of original companion app of the ZeTime are used to it. In the app of MyKronoz (manufacturer of the ZeTime watch) when you open a preference menu, the values are updated off the watch before you could change anything.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/zetime/ZeTimePreferenceActivity.java
cpfeiffer: No offense intended, you do a great job! We can probably make use of this method elsewhere too, maybe with an additional argument for the string for onSendConfiguration().

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/zetime/ZeTimeDeviceSupport.java
cpfeiffer: We will soon see if it is related to DST :-)

cpfeiffer:
You could also retrieve the configuration during or after initialization, so the preferences activity would not need to be updated. Does the watch send a notification when a preference is changed on the watch, so that you could sync the value to Gadgetbrige immediately? Otherwise we would still need to sync when opening the preferences.

Lightwars:
I can do that, thought about this before.
I tested if we get a change notification, but there was none.

cpfeiffer:
Excellent, great job! :confetti_ball: :+1:

PR #1339, opened 2018-11-17
Avoid KIA Sportage appearing in the list of devices

sgomin:
Don't display KIA Sportage in the list of devices


cpfeiffer:
Checking for startsWith("SPORT") is certainly not a polite thing :face_with_head_bandage:
For lack of a better detection, I will accept your PR; for the long run, we should find a better way to detect the desired device.

Thanks!

Issue #1344, opened 2018-11-24
[Mi Band 3] only Telegram notifications display unknown characters `?`


matttbe:
First, thank you for developing and maintaining this excellent application! Without Gadgetbridge,  I am not sure I would have allowed a proprietary application to collect personal data!
I was an happy user with my Mi Band 2 and I am now testing the Mi Band 3.

#### Before reporting a bug, please confirm the following:
- [X] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [X] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [X] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
When I receive Telegram notifications, a lot of `?` characters are displayed before showing the content of the notification:

![img_20181124_085817_1](https://user-images.githubusercontent.com/768677/48966246-f0145a80-efcd-11e8-8f36-847de8263dea.jpg)

Here are the logs generated by the application when I received this notification shown in previous photo:

```java
11-24 08:56:50.590 I/et.gadgetbridge(20875): type=1400 audit(0.0:9326): avc: denied { open } for uid=10192 path="/proc/vmstat" dev="proc" ino=4026532375 scontext=u:r:untrusted_app_27:s0:c512,c768 tcontext=u:object_r:proc_vmstat:s0 tclass=file permissive=1
11-24 08:56:50.590 I/et.gadgetbridge(20875): type=1400 audit(0.0:9327): avc: denied { getattr } for uid=10192 path="/proc/vmstat" dev="proc" ino=4026532375 scontext=u:r:untrusted_app_27:s0:c512,c768 tcontext=u:object_r:proc_vmstat:s0 tclass=file permissive=1
11-24 08:56:53.644 I/nodomain.freeyourgadget.gadgetbridge.externalevents.NotificationListener(20875): Processing notification 1543043049 age: 2644 from source org.telegram.messenger with flags: 16
11-24 08:56:53.644 I/nodomain.freeyourgadget.gadgetbridge.externalevents.NotificationListener(20875): found wearable action: 2 - Reply to Motorola | ArrowOS  null
[...]

PR #1346, opened 2018-11-25
notif: sanitize strings set by external apps

matttbe:
It seems some apps like Telegram add a lot of Unicode Control sequences
for unknown reasons. Because these strings are set by external apps, it
is safer to sanitize them. For the moment, only Unicode control
sequences are stripped.

Fixes: #1344


ashimokawa:
@roigreenberg 
Do you think this would break something?

roigreenberg:
I can't see any reason why it will change anything.

PR #1347, opened 2018-11-25
notif: convert emoji's to supported chars

matttbe:
messages: convert emoji's to supported chars

Some devices don't support emoji's and display a series of `?` chars.
Instead of that, replace them by `:<emoji>:` text, e.g. `A 🐱` is
replaced by `A :cat:`.

This is done by using 'android-emojify' project and their
`EmojiUtils.shortCodify(text);` function.

For more details about this lib released under MIT license:

    https://anitrend.github.io/android-emojify/

Note that for the moment, only amazfitbip's devices have this feature
because I don't know if the other watches can display Emoji. Simply use
the function mentioned here above to add this feature for other devices.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
@@ -53,6 +53,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import io.wax911.emojify.EmojiManager;
 import nodomain.freeyourgadget.gadgetbridge.database.DBHandler;
 import nodomain.freeyourgadget.gadgetbridge.database.DBHelper;
 import nodomain.freeyourgadget.gadgetbridge.database.DBOpenHelper;
@@ -197,6 +198,8 @@ public class GBApplication extends Application {
             }
             startService(new Intent(this, NotificationCollectorMonitorService.class));
         }
+
+        EmojiManager.initEmojiData(getApplicationContext());
     }
 
     @Override
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -87,6 +87,7 @@ dependencies {
     implementation "org.cyanogenmod:platform.sdk:6.0"
     implementation 'com.jaredrummler:colorpicker:1.0.2'
 //    implementation project(":DaoCore")
+    implementation 'com.github.wax911:android-emojify:0.1.7'
 }
 
 preBuild.dependsOn(":GBDaoGenerator:genSources")
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractDeviceCoordinator.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/AbstractDeviceCoordinator.java
@@ -149,4 +149,7 @@ public abstract class AbstractDeviceCoordinator implements DeviceCoordinator {
     public int[] getColorPresets() {
         return new int[0];
     }
+
+    @Override
+    public boolean supportsUnicodeEmojis() { return false; }
 }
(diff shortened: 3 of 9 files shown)


ashimokawa:
@matttbe 
Cool.

But this cant be merged as is, since the Amazfit Cor does support some emoji and runs off AmazfitBipSupport also.

I think we need some getter in *Coordinator classes, and do this in generic code while checking for emoji support on a per-device basis.

AFAIK two devices have emoji support: Pebble and Amazfit Cor. At least the Pebble can have custom fonts with more emoji, Amazfit Cor has only a few that are top-ranked in China IIRC.

matttbe:
> But this cant be merged as is, since the Amazfit Cor does support some emoji and runs off AmazfitBipSupport also.

Good to know, thank you for the info :-)

> I think we need some getter in *Coordinator classes, and do this in generic code while checking for emoji support on a per-device basis.

This info should indeed be put "higher" in the code architecture. Then I guess you would like have something like this in AmazfitBipSupport class:

```java
    if (gbDevice.isEmojiSupported())
        message = StringUtils.convertEmoji(message);
```

or only:

```java
    message = gbDevice.sanitizeEmoji(message);
```

> AFAIK two devices have emoji support: Pebble and Amazfit Cor. At least the Pebble can have custom fonts with more emoji, Amazfit Cor has only a few that are top-ranked in China IIRC.

OK, I can try to add support for other devices :)

matttbe:
@ashimokawa : I just updated the PR.

Regarding https://github.com/Freeyourgadget/Gadgetbridge/pull/1347/commits/1a423a519f9fb2cebb4ab9c7f302841d8a0167cc , was it the kind of code you had in mind?

ashimokawa:
I think this should be done in generic code, not in AmazfitBipSupport, so all (future) devices can benefit from the code just by implementing supportsUnicodeEmojis() and returning true.

matttbe:
That would be good but I don't how to do that with the current code. If we look at `onNotification()` function in `AmazfitBipSupport` class, we create the message to display depending of  the device using data coming from `notificationSpec` and we end by adding an array of bytes in a queue.

So for the moment, each class implementing `DeviceSupport` have to handle the modification of the message separately, no? Except if at the end of `onNotification()`, devices that need to display something on a watch have to give the message (String type) to a function that can then do transformations and send it to the builder queue. But I don't know if it is possible to do that in a generic way for all devices. I guess they don't all use the same command. Then we would need another function specific for a device to build the command bit from a message that had been "sanitized".

What do you think?

It is just that I don't know the other devices. I am not sure how to have a generic way to do that :-)

matttbe:
> The only location where all receivers send their notification is in DeviceCommunicationService.handleAction()

Good idea, thank you!
What do you think about the new version? :-)

matttbe:
Hi @ashimokawa

Is there anything I can do to help merging this? :-)

Merry Xmas!

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
```diff
@@ -197,6 +198,8 @@ public void onCreate() {
             }
             startService(new Intent(this, NotificationCollectorMonitorService.class));
         }
+
+        EmojiManager.initEmojiData(getApplicationContext());
```
cpfeiffer: I would like to see this be done on demand, e.g. in sanitizeNotifText() so that this will not increase the startup time. Additional benefit would be that we only need to do it when really needed. 

It depends on whether initEmojiData() is callable from a background thread, though.

cpfeiffer:
So to move forward, I'm merging this as is, and would be happy if you could try to move the EmojiManager initialization somewhere to a lazy place :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
matttbe: That's a good idea! I just tested that and it seems it is working fine!
(at the end, this function is reading a file and parsing it, should be fine in a background thread, no?)

matttbe:
Thank you for having already merged this change!

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/GBApplication.java
cpfeiffer: Yes, that should be fine. I'm just always wary of calling "foreign" APIs from background threads, especially when it comes to initialization, loading of resources (icons) etc.

PR #1371, opened 2018-12-22
Pebble M7S Watchface: add weather support 

jcrode:
This patch adds rudimentary support for weather information on the M7S pebble watchface (https://github.com/cfg1/pebble-mss). It lacks support for user-configurable things like customizable weather strings / units.


cpfeiffer:
Nice work, looks good to me, but I'll let @ashimokawa or @danielegobbetti comment, because Pebble's are mostly their area of expertise.

PR #1372, opened 2018-12-22
[Chart] change data shown on diagrams - steps or activity

Dikay900:
an additional setting under chart settings to switch between steps and activity to be shown on all diagrams


cpfeiffer:
Thanks a lot for your contribution! I have yet to try out the change to see how it looks. Did the removal of the maximum y axis value have any noticeable effect on how the activity chart looks?

PR #1373, opened 2018-12-23
emoji: do a lazy initialisation

matttbe:
The goal is not to slowdown the startup but also to do this initialisation only when it is needed.

Suggested-by: @cpfeiffer 

Note for @cpfeiffer : in your suggestion, you said:

> I would like to see this be done on demand, e.g. in `sanitizeNotifText()`

I moved this to the new EmojiConverter class to do everything related to the emoji there. I hope it  is OK for you :)

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/EmojiConverter.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/EmojiConverter.java
@@ -17,6 +17,9 @@
 
 package nodomain.freeyourgadget.gadgetbridge.util;
 
+import android.content.Context;
+
+import io.wax911.emojify.EmojiManager;
 import io.wax911.emojify.EmojiUtils;
 
 public class EmojiConverter {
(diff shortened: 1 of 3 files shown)


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/util/EmojiConverter.java
```diff
@@ -56,17 +61,30 @@
...
+
     private static String convertSimpleEmojiToAscii(String text) {
         for (String[] emojiMap : simpleEmojiMapping) {
             text = text.replace(emojiMap[0], emojiMap[1]);
         }
         return text;
     }
 
-    public static String convertUnicodeEmojiToAscii(String text) {
+    private static String convertAdvancedEmojiToAscii(String text, Context context) {
```
cpfeiffer:
```suggestion
     private static synchronized String convertAdvancedEmojiToAscii(String text, Context context) {
```
to prevent concurrency issues in case this is ever called from multiple threads in the future.

cpfeiffer:
Thanks :+1:  :-)

PR #1374, opened 2018-12-24
emoji: only the init can be sync

matttbe:
After having saw @cpfeiffer 's suggestion, I did this change, then I saw it was already done :)

Here only the init part is set as 'synchronized', very small optimisation but because this change is ready, why not sharing it :)


cpfeiffer:
Good, thanks :+1:

Issue #1393, opened 2019-01-14
Notify about Calls from VoIP Apps


veecue:
#### Your issue is:
I would like to get call notifications about VoIP calls via IMs (e.g. Telegram, Threema, WhatsApp) and via specialized Apps (like FRITZ! Box Fon). A bit of research shows, that most of those apps Send an ongoing notification with an Accept and Reject PendingIntent that could be parsed. Telegram however only does so, when the device is unlocked and displays a custom screen when the device is locked. Any ideas on how to solve this?

#### Your wearable device is:
Mi Band 3

*Please specify model and firmware version if possible*

#### Your android version is:
7.0

#### Your Gadgetbridge version is:
master

*New requests about already solved/documented topics could be closed without further comments. Same for too generic or incomplete reports.*

ashimokawa:
Yep. Agree. Problem is that when scraping button captions we have to support tons of languages. 😒

Or is there a way to get untranslated strings? I guess no ...

veecue:
I think that parsing the button order would be the most reliable thing to do (per app :(). Do you have an idea on how to get something working with Telegram? I will start writing an implementation that catches ongoing Notifications of type "call" and tries to extract the reject button's Intent (the mi band 3 can't accept, so for now, I'll start with reject)

MMurzik:
I'd like call notification from Zoiper pro VoIP app too

veecue:
@MMurzik can you check out my proof of concept code (declining doesn't work yet) and test it with your app? Would be cool if you could submit the debug output aswell

PR #1396, opened 2019-01-19
WIP: added nofications for VoIP calls

veecue:


veecue:
will fix #1393

cpfeiffer:
This is pretty cool! With which apps did you test it? @ashimokawa tried with Signal, but it didn't work for him.

veecue:
I've tested it with:
 * Telegram (only works when screen is unlocked)
 * Threema (works well, even displays caller)
 * Whatsapp (same as Threema)
 * FRITZ! App Fon (notifies on call, stops when call is rejected, but keeps vibrating when call is accepted)

As already stated in #1393: The current implementation is just a PoC and doesn't support rejecting or accepting calls yet.

I will continue developement when you are willing to have the current concept (reading out notifications, hardcoded per-app settings (e.g. Accept/Reject button order) in the app.

ashimokawa:
@veecue 
I don not have a better idea and I also want that feature ;), so hard-coding stuff would be okay for me, as long it it does not assume your phone being set to English...

cpfeiffer:
Thanks a bunch, looking forward to further refinements :+1: :+1:  :smiley:

PR #1404, opened 2019-01-29
Add initial support for Gatt server in BtLEQueue

andyboeh:
As discussed with @cpfeiffer in #1377, this PR adds generic support for GATT server operations in the BtLEQueue.

I tested it only very briefly on the Casio GB-5600B (relevant code is not part of the PR), where it seems to work well. However, I'm posting it anyway to hear your feedback.

For a device to use it, the device has to set up the `BluetoothGattServer` in the constructor, similar to the calls to `addSupportedService`. It's a bit more complex since the characteristics and their properties need to be configured first, though.

Similar to read/write requests, the device has to override the `onCharacteristicReadRequest` method and can use a call to `performServer` to generate a new `ServerTransaction`. Currently, only responding to read requests is implemented since this is the only operation the Casio watches perform.


cpfeiffer:
Thanks a lot! Reviewing the other PR took a little longer, so I will have to defer this one. Just one question in advance: do you think it would be possible to have the implementation in two separate classes (one for the client and one for the server)? They might share part of the implementation through a shared superclass of course, but would it be possible to have to separate instance of them, operating in two separate threads?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/BtLEQueue.java
```diff
@@ -53,20 +57,27 @@
...
+    private final Object mTransactionMonitor = new Object();
     private final GBDevice mGbDevice;
     private final BluetoothAdapter mBluetoothAdapter;
     private BluetoothGatt mBluetoothGatt;
+    private BluetoothGattServer mBluetoothGattServer;
+    private final Set<BluetoothGattService> mSupportedServerServices;
 
-    private final BlockingQueue<Transaction> mTransactions = new LinkedBlockingQueue<>();
+    private final Queue<Transaction> mTransactions = new ConcurrentLinkedQueue<>();
+    private final Queue<ServerTransaction> mServerTransactions = new ConcurrentLinkedQueue<>();
```
cpfeiffer: Have you considered using a single queue for both client and server transactions? We could keep using the LinkedBlockingQueue, spare the mTransactionMonitor, the separate wait() and notify().

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/BtLEQueue.java
```diff
@@ -77,7 +88,17 @@ public void run() {
 
             while (!mDisposed && !mCrashed) {
                 try {
-                    Transaction transaction = mTransactions.take();
+                    if(mTransactions.isEmpty() && mServerTransactions.isEmpty()) {
+                        synchronized (mTransactionMonitor) {
+                            try {
+                                mTransactionMonitor.wait();
+                            } catch (InterruptedException e) {
+                                e.printStackTrace();
```
cpfeiffer: We should either ignore or log this exception (if we keep mTransactionMonitor).

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/BtLEQueue.java
andyboeh: Yes, that was my original plan. However, I did not succeed because, at a later stage, I couldn't distinguish between client and server operations anymore and ended up with a mess. I can give it a second try, though.

andyboeh: OK, I took another approach now and I'm back to one queue and the old queue handling stuff.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/AbstractTransaction.java
```diff
@@ -0,0 +1,49 @@
...
+    public String getTaskName() {
+        return mName;
+    }
+
+    protected String getCreationTime() {
+        return DateFormat.getTimeInstance(DateFormat.MEDIUM).format(new Date(creationTimestamp));
+    }
+
+    public int getActionSize() {
+        return 0;
```
cpfeiffer: This method should rather be abstract (and named getActionCount(), but that might just be my personal preference :-)

Apart from that: :+1: :+1:

cpfeiffer:
Great job! I added back CasioGATTServer.java since it is still in use while your other PRs arenot merged.

I hope we csn rebase the other PRs on top of this one, or reset and cherry-pick the additional commits. I will install and test this version now.

Thanks!

cpfeiffer:
I think there's an issue left with `mWaitForServerActionResultLatch` in `BtLEQueue`. The field is never initialized and there's no appropriate countdown() call.

andyboeh:
You are right - I copied that part incompletely from the client actions. I never run into issues since the only action currently defined does not expect a result. For now, it could be removed, but I suppose it's better to implement it completely.
Shall I create a separate PR or will you fix it?

cpfeiffer:
I can do that. AFAICS we could even remove the field and reuse the other one.

cpfeiffer:
When looking at this, I think I got a slightly better picture of this issue. AFAIU, these are our requirements:
- have all BLE communication in the same thread 
    - (using BtLEQueue, done)
- server actions should be run ASAP upon request
    - could use BtLEQueue#insert() (performImmediately())
- send notifications on request (no immediate need)
    - this is where we probably need a queue with support for waiting (for onNotificationSent)

Please add, if you have other things in mind.

PR #1408, opened 2019-01-31
Add initial support for Xiaomi Mi Scale 2

jfgreffier:
Initial support for Mi Scale 2 aka Mi Body Composition Scale : MIBCS discovery, immediate measure of weight with Date and unit.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miscale2/MiScale2DeviceSupport.java
```diff
@@ -0,0 +1,290 @@
...
+        LOG.warn("Device info: " + info);
+        versionCmd.hwVersion = info.getHardwareRevision();
+        versionCmd.fwVersion = info.getSoftwareRevision();
+        handleGBDeviceEvent(versionCmd);
+    }
+
+    private void handleWeightInfo(Date date, float weight, String unit) {
+        // TODO
+        LOG.warn("Weight info: " + weight + unit);
+        GB.toast(weight + unit, Toast.LENGTH_SHORT, GB.INFO);
```
cpfeiffer: Here we need to think how to handle this UI-wise and data-wise, I guess.

We should probably store the results in the db. And fire an event so that the UI can display the current value and possibly a chart with old values.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miscale2/MiScale2DeviceSupport.java
```diff
@@ -0,0 +1,290 @@
...
+
+    @Override
+    public boolean onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteristic characteristic) {
+        super.onCharacteristicChanged(gatt, characteristic);
+
+        UUID characteristicUUID = characteristic.getUuid();
+        if (characteristicUUID.equals(GattCharacteristic.UUID_CHARACTERISTIC_BODY_COMPOSITION_MEASUREMENT)) {
+            final byte[] data = characteristic.getValue();
+
+            boolean stabilized = testBit(data[1], 5) && !testBit(data[1], 7);
```
cpfeiffer: Does this implementation conform to the standard BLE profile/service for this? If so, we could even implement it generically like the DeviceInfoProfile, so that it can be reused for other devices.

build.gradle
```diff
@@ -15,10 +15,10 @@ buildscript {
 
 allprojects {
     repositories {
-        jcenter()
         maven {
             url "https://jitpack.io"
         }
         google()
+        jcenter()
```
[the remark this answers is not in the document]
jfgreffier: I've met this issue at build
https://issuetracker.google.com/issues/80362794

I can't reproduce it now, we can probably remove this change

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miscale2/MiScale2DeviceSupport.java
jfgreffier: Unfortunately, no, it is not standard.
Here is the spec:
https://www.bluetooth.com/specifications/gatt/viewer?attributeXmlFile=org.bluetooth.characteristic.body_composition_measurement.xml

The order of the data is quite different

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/miscale2/MiScale2DeviceSupport.java
jfgreffier: The amount of changes were a bit overwhelming to me :)
Yes, store a body information in DB is probably the next step

build.gradle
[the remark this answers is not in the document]
jfgreffier: Reverted.

cpfeiffer:
Alright, I merged this in, we can add the db and UI stuff bit by bit.

Thanks a lot! :smile:

PR #1416, opened 2019-02-05
Add support for ble scanning to speed up reconnection

andyboeh:
As discussed in #1399 this adds support for BLE background scanning to speed up automatic BLE reconnect. Since this uses more power than the automatic reconnect, it needs to be configured on a per-device basis. 

This work is based on improved BtLEQueue with Gatt server handling features, so #1404 should be merged first. I'm posting it early to get feedback.


andyboeh:
Testing has shown that a bug that prevented the scan from being stopped leads to excessive battery usage. This should be fixed now in two ways:
  - after five minutes of low latency scanning it switches to balanced scanning
  - scanning is successfully stopped if the wanted device has been found

Unfortunately, setting a ScanFilter with the device address of the previously connected device does not work, results are only received if another app scans at the same time. Thus, the filtering is performed in the BluetoothScanCallbackReceiver

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/BtLEQueue.java
```diff
@@ -252,16 +370,107 @@ private void handleDisconnected(int status) {
...
+        }
+
+        @Override
+        public void onScanFailed(int errorCode) {
+            LOG.error("ScanCallback.onScanFailed:" + errorCode);
+        }
+    };
+
     public void dispose() {
         if (mDisposed) {
```
cpfeiffer: I think the scanning and all those runnables need to be stopped/cleared on explicit `connect()` or `disconnect()`. Or is this taken care of already?

app/src/main/AndroidManifest.xml
```diff
@@ -338,6 +338,13 @@
                 <action android:name="android.bluetooth.adapter.action.STATE_CHANGED" />
             </intent-filter>
         </receiver>
+        <receiver
```
cpfeiffer: What's the lifecycle of this receiver? Is it running all the time, basically as a singleton? Is that intended? What if there were two devices, waiting for reconnection?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothScanCallbackReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+        String uuid = intent.getExtras().getString("uuid");
+
+        int bleCallbackType = intent.getIntExtra(BluetoothLeScanner.EXTRA_CALLBACK_TYPE, -1);
+        if(bleCallbackType != -1) {
+            //LOG.debug("Passive background scan callback type: " + bleCallbackType);
+            ArrayList<ScanResult> scanResults = intent.getParcelableArrayListExtra(BluetoothLeScanner.EXTRA_LIST_SCAN_RESULT);
+            for(ScanResult result: scanResults) {
+                BluetoothDevice device = result.getDevice();
+                if(device.getAddress().equals(wantedAddress) && !mSeenScanCallbackUUIDs.contains(uuid)) {
+                    mSeenScanCallbackUUIDs.add(uuid);
```
cpfeiffer: This list is never cleared, AFAICS. Also, maybe a HashSet would be more appropriate, since the receiver may be listening for a long time.

andyboeh:
Thank you very much for the thorough review! TBH, I fixed a few of your comments already locally, but haven't had the time to push it, yet. I will look through your comments during the next days and see what needs to be done.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/AbstractDeviceSupport.java
```diff
@@ -142,6 +142,11 @@ public Context getContext() {
         return context;
     }
 
+    @Override
+    public boolean useBleScannerForReconnect() {
+        return false;
```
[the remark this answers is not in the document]
andyboeh: You mean similar to "Enable automatic BLE reconnect", globally for all devices? I was thinking on a per-device configuration, since the Casio devices require this anyway.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothScanCallbackReceiver.java
```diff
@@ -0,0 +1,78 @@
...
+                    LOG.info("ScanCallbackReceiver has found " + device.getAddress() + "(" + device.getName() + ")");
+                    BluetoothAdapter.getDefaultAdapter().getBluetoothLeScanner().stopScan(getScanCallbackIntent(GBApplication.getContext(), wantedAddress, uuid));
+                    GBApplication.deviceService().connect();
```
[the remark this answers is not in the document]
andyboeh: TBH I have no clue on how to get the GBDevice in my BroadcastReceiver. The method here completely disposes the old device service and creates a new one.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/btle/BtLEQueue.java
andyboeh: Nope, you are right. I'm currently testing a fix for this (I ran into this problem already).

app/src/main/AndroidManifest.xml
andyboeh: Does GB support multiple devices in a waiting for reconnection state already? 
As far as I have understood the Android-concept here, the receiver is always active and acting like a singleton. Since the address to match and a uuid are given as extras to the intent when starting a scan, it shouldn't be a problem to have several devices in a waiting for reconnect state.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothScanCallbackReceiver.java
andyboeh: True, I haven't yet figured out the best point in time / the best way to clear this list. Without it, I end up with several connect requests since the device is found multiple times.

app/src/main/AndroidManifest.xml
cpfeiffer: No, not yet, I'm just checking for the future. There's even a WIP PR for multi device PR, but we haven't gotten around at making it ready for prime time. :-(

I think I understand the intention now, maybe deserves a javadoc explanation :-)

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothScanCallbackReceiver.java
cpfeiffer: When scanning has stopped, I would say.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/BluetoothScanCallbackReceiver.java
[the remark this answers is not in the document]
cpfeiffer: See `DeviceHelper.toSupportedDevice(BluetoothDevice)`

Where does it dispose the device service?

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/AbstractDeviceSupport.java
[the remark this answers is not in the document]
cpfeiffer: The default implementation could check the global configuration and the casio implemention could override to forefully enable it.

I agree a pretty device-specific thing. I think until we have better device specific configurabilities, we could make it a global option.

andyboeh:
I think I got all the comments fixed now - regarding the UUID in the ScanCallbackReceiver: The purpose is to know when a device has been found for a given scan. The receiver is called several times for the same device and it might get called even after stopping the scan. IMHO it is sufficient to remember the last UUID that was scanned, so I removed the List and replaced it by a simple String. For multi-device-support, this probably needs fixing anyway, since there is no callback when scanning has really stopped.

andyboeh:
Rebasing was not exactly easy, so I merged everything into one single commit and pushed that. I hope I got everything alright.

andyboeh:
@cpfeiffer any updates on this?

andyboeh:
I'm closing this in favor of #1502 where feedback from code review has been implemented, but the dependency on #1416 is not present.

PR #1423, opened 2019-02-11
Added Foreground Service permission

TaaviE:
[This permission](https://developer.android.com/reference/android/Manifest.permission#FOREGROUND_SERVICE) [will be required](https://developer.android.com/about/versions/pie/android-9.0-migration#tya) if targetSDK is bumped, it doesn't hurt to make this change preemptively.


TaaviE:
Travis build failed because Greendao (1998d7cd2d21f662c6044f6ccf3b3a251bbad341) is somehow not found.

cpfeiffer:
Yes, totally makes sense, thanks! :+1:

PR #1435, opened 2019-02-17
Tasker Plugin

christian-n:
# Tasker Plugin

### Features
- Tasks over Intents
- Task Service for scheduling
- Event based tasks with threshold
- Full-fledged user settings
- Easy to implement

### Implemeted devices
- XWatch

### Quickstart
To implement a device add a `TaskerDevice` like
```java
    XWATCH(DeviceType.XWATCH, new XWatchTaskerSpec(DeviceType.XWATCH));
```
all you need is a `DeviceType` and a `TaskerSpec`. Last thing is to add the `TaskerBleProfile` to your `AbstractBTLEDeviceSupport` like 
```java
    addSupportedProfile(new TaskerBleProfile<>(this, TaskerDevice.XWATCH));
```
> Note
> If you want generic user settings you MUST extend `AbstractTaskerSpec` or atleast use `PreferenceTaskerSettings` as `TaskerSettings`.

### Tasker Hook
Tasker hook is done with `TaskerBleProfile` as an `AbstractBleProfile` (correct me if this is not the best way).

### Tasker Service
The `AbstractTaskerService` is used to schedule `TaskerTask`'s. The *threshold* is used to cumulate calls. The `TaskerEvent` is then send with the specific amount of calls within that *threshold*.
The default service for the hook is `SpecTaskerService`.

It uses the `TaskerSpec`'s `TaskerSettings` as configuration. There is a `TaskerService` implementation for programmatic use.

### Tasker Event
The default `TaskerEventType`'s are
- Button: For button related events.
- Connection: For bluetooth connection related events.
- Data: For data receiving related events.
- No_op: For non operational events.

You can easly add more types via `TaskerEventType` class. It does not effect the already implemented devices.
Don't forget to add your type to
```java
    public static List<TaskerEventType> getTypes() {
        return Arrays.asList(BUTTON, CONNECTION, DATA);
    }
```
And for device support add it to the device's `TaskerSpec`
```java
    List<TaskerEventType> getSupportedTypes();
```
### Tasker Settings
Tasker user settings are implemented via
##### TaskerActivity
As main entry point from `SettingsActivity`. It lists `TaskerDevice`'s and has a *enabled* preference for overall tasker support.
Forwards to `TaskerEventsActivity`.

Lists all event types from `TaskerSpec`'s `getSupportedTypes()` method.
Forwards to `TaskerEventActivity`.

Has preferences for
- Enabling the events
- Enabled the threshold
- Setting the threshold
- Setting a task name
- Adding/Removing more tasks for threshold calls (like double button press)

### ToDo
- More devices should be implemented.
- Only EN and DE is currently translated.

### Commit Information
- Includes JavaDoc


cpfeiffer:
Wow, that's very nice! I will review this (and some other PRs) tomorrow.

IzzySoft:
Though this might bring us the NonFreeAdd AntiFeature ("promoting" a non free app) at F-Droid, I can't wait to see this 😃

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/xwatch/XWatchTaskerSpec.java
```diff
@@ -0,0 +1,41 @@
...
+        if (XWatchService.UUID_NOTIFY.equals(characteristic.getUuid())) {
+            byte[] data = characteristic.getValue();
+            if (data[0] == XWatchService.COMMAND_ACTIVITY_DATA) {
+                return TaskerEventType.DATA;
+            }
+            if (data[0] == XWatchService.COMMAND_ACTION_BUTTON) {
+                return TaskerEventType.BUTTON;
+            }
+            if (data[0] == XWatchService.COMMAND_CONNECTED) {
+                return TaskerEventType.CONNECTION;
```
cpfeiffer: I think that such events would rather be handled generically via GBDevice#getState() changes (see GBDevice#sendDeviceUpdateEvent()).

christian-n:
I will be on it tomorrow. Thanks for the review!

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/xwatch/XWatchTaskerSpec.java
christian-n: Can you give me more info on that.

christian-n:
@cpfeiffer Ive done some of your fixes but i dont know what to do with getState.
Seems no one is using it. at least ive looked up the XWatchSupport and MiBandSupport.

cpfeiffer:
Thanks for the updates!

GBDevice#getState() is used all over the place. It's the state describing whether a device is disconnected, connected, initializing, busy, etc. I thought that these state changes would be good candidates for tasker interaction, especially since they could be handled in a device independent way.

The *Support-classes usually only *set* the state, but don't react to any changes.

cpfeiffer:
Please let me know if I should explain in more details.

PR #1453, opened 2019-03-12
Initial implementation for Mi Band 2 'custom menu'

mvallerie:
Hi,

First of all, thanks for this great app.

This PR implements a "custom menu" feature using the func button of the Mi Band 2. Something very similar to what is shown in [this YouTube video](https://www.youtube.com/watch?v=i9QoAukmoGs).

Here are the needed configuration steps : 
- Enable the Mi Band Action button
- Set a minimum value for "Delay Button After Action" (otherwise, the first menu option may not display correctly on the band).
- Enable the Custom Menu under "Button Action" > "Custom Menu" and set the "Menu Elements" (that's very gross for now...you have to input it as `Label1:Message1;Label2:Message2;...`).
- Enable Text notifications in Mi Band 2 specific settings
- You're good to go : now, the "Button Action" will trigger the custom menu opening. Then, single tap to move forward in the menu, double tap to move backwards, and triple tap to validate (so it will broadcast the associated message, which you can catch in any automation tool). If you don't do anything during 5 secs, the menu will close.

It's working great but : 
- The band vibrates every time it gets a "menu notification", i'm not sure why, and i couldn't find a way to get around.
- I have no idea if this PR may or may not have impact on other existing devices support. It should not, but you never know. I hope i'm not adding more spaghettis :-).

Some improveable-s :
- Custom separators / better UI for the "Menu Elements" string
- Custom controls for the menu (easy one, but i don't have much time this week).

Thanks in adv. for your feedback.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBandConst.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/miband/MiBandConst.java
@@ -32,6 +32,8 @@ public final class MiBandConst {
     public static final String PREF_MIBAND_ALARMS = "mi_alarms";
     public static final String PREF_MIBAND_DONT_ACK_TRANSFER = "mi_dont_ack_transfer";
     public static final String PREF_MIBAND_RESERVE_ALARM_FOR_CALENDAR = "mi_reserve_alarm_calendar";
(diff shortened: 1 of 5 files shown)


mvallerie:
I added configurable menu controls.

As said in my edit, the only issue i had so far using this is when the screen is off. Sometimes the menu becomes unresponsive (ie : the phone doesn't send the alert to the band immediately, there's a 5-10 secs delay).

PR #1456, opened 2019-03-16
Fixes GPX export namespace declarations

nickspacek:
I noticed that the XML being output from the GPXExporter was broken in that elements did not belong to the proper namespaces. The crux of the problem was that the default namespace was not set to the GPX namespace.

I renamed some of the constants to make it more clear what was being used. I also removed the use of the namespace when adding attributes; as per the GPX schema: "By default, local attribute declarations have no namespace."

I added some really basic tests using some sample data I collected. The tests create two GPX files, one with heart rate data and one without, and run schema validation against them. Prior to the GPXExporter changes I introduced, validation failed.

It's possible that depending on the strictness of the parser, they were able to deal with the incorrect namespaces (e.g. by only looking at tag names). This seems to be the case with Osmand imports. However, strict/proper parsing of the XML would fail when namespaces were considered. The output files are now understood by Garmin Connect in my tests, where the original exports were failing.


cpfeiffer:
Thanks a lot! :heart: :ok_hand:

PR #1492, opened 2019-04-23
Add pebblekit extension for reopening last app

matejdro:
There are several 3rd party Pebble apps that open up momentarily via PebbleKit (for example to pop up alert or notification). Those app start themselves via PebbleKit start and stop intents.

However, there is slight usability issue with those: If you have anything else open on the watch other than watchface (for example a running timer or sports tracker) when momentary app opens, previous app will be closed and after new app is finished, watch will return to the watchface. That can be annoying to the user.

With official Pebble app, this problem was solved by hooking into developer connection service and sniffing packets from there. It was very hacky and insecure. With Gadgetbridge, there is no developer connection, but since this app is open source, I figure I could add extension to the PebbleKit interface that would enable this functionality in much saner way.

PR adds `com.getpebble.action.app.REOPEN_LAST_APP` extra to the `com.getpebble.action.app.STOP` PebbleKit intent. If extra is true, app will be replaced by previously running app instead of watchface.

### DIFF
--- a/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleActiveAppTracker.java
+++ b/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/pebble/PebbleActiveAppTracker.java
@@ -1,0 +1,40 @@
+package nodomain.freeyourgadget.gadgetbridge.service.devices.pebble;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import java.util.UUID;
+
+public class PebbleActiveAppTracker {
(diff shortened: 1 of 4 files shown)


ashimokawa:
Cool. @danielegobbetti, any objections merging this?

danielegobbetti:
@ashimokawa if it looks good to you, it looks good to me :)

ashimokawa:
@matejdro 
Do you want this merged or do you want to drop this in favor of #1495 ?
We want to release on Thursday -  just tell us.

matejdro:
I don't see a reason to not merge this in #1495 will take some time, so this allows apps to already use the new functionality until then.

PR #1502, opened 2019-05-01
Improve Casio device support without BLE scanner

andyboeh:
Since #1416 has not been accepted yet, I pushed all the updates to the Casio device support to a new branch. The following has changed:

- Fix setting alarms by providing F/W version
- Make use of the BtLE server queue
- Move initialization to a separate InitOperation
- Improve connection process
- Improve pairing process by supplying a connectFirstTime() method
- Produce less connection attempts by not allowing certain operations when device is not connected (it does not make sense to try connecting to the watch, since the watch is only in connectable state for a few seconds)

I hope that this PR can be merged quicker since it only affects the Casio watch.


ashimokawa:
Thanks :)

PR #1503, opened 2019-05-01
Also check work profile for number lookup

andyboeh:
If a device contains a work profile and a contact saved in the work profile is calling, the name is not queried within Gadgetbridge. This PR also checks the work profile for matching contacts in case the number is not found.


ashimokawa:
Are you suse ENTERPRISE_CONTENT_FILTER_URI will look up in both profiles?

andyboeh:
According to the docs and to my tests it does so :) 

>  If the device has no corp profile that is linked to the current profile, it behaves in the exact same way as CONTENT_FILTER_URI. If there is a corp profile linked to the current profile, it first queries against the personal contact database, and if no matching contacts are found there, then queries against the corp contacts database. 

See here for reference: https://developer.android.com/reference/android/provider/ContactsContract.PhoneLookup#ENTERPRISE_CONTENT_FILTER_URI

PR #1515, opened 2019-05-14
Amazfit Bip: Add new latin languages

DerFetzer:
Since Latin firmware 1.1.5.36 there are additional languages available:
- German
- Italian
- French
- Turkish

I tested switching between them without any problems.


ashimokawa:
Thanks!

PR #1523, opened 2019-05-25
Add support for ignoring (muting) calls

andyboeh:
Muting an incoming call is an important feature for me which, unfortunately, Gadgetbridge is lacking. This PR adds support for it by temporarily changing the ringerMode to silent and restoring it back once the call has finished.

Although there are "silenceRinger" methods in ITelephony and TelecomManager, they are only available to system applications. Hence, the workaround is necessary.

The Casio implementation has been updated to add support for ignoring calls.


ashimokawa:
Thanks!

PR #1560, opened 2019-07-17
NEW FEATURE: Add preference to turn your clock forward or backward

Lightwars:
Some people might like this one...
Back in the days of "normal" clocks some liked to turn their clock forward by some minutes. They did it to be not too late for appointments or just to have some extra time.

So to get our gadgets to be ahead of our local time, I introduce a new preference and functions which gives back a calendar or date object, which is set according to that preference.
You can also set your gadget time behind your local time by some minutes, maybe someone likes this as well.

I have tested this with my ZeTime and it works as intended. But I do not own any other device to test the functionality thoroughly.


Lightwars:
There is a draw back of playing with the time of the gadget, all activity data timestamps will have an offset.

I would be best to have a voting, if such a feature is wanted by the people.

PR #1568, opened 2019-07-25
Formatted device information

dakhnod:
Made proprietary Device information show up like "INFO: abc" instead of "INFOabc".

That problem could be avoided by using "INFO: " as the information key/name, but that seems to miss the point of a hashlist for me...


dakhnod:
wait, that is not the move...

i just think that code like 

gbDevice.addDeviceInfo(new GenericItem("SUPPORTS_VIBRATION: ", String.valueOf(supportsExtendedVibration())));

just doesn't feel right... 

would it be reasonable to add an extra field named Lebel, where one could put the actual Label to display?

PR #1571, opened 2019-07-26
Fixed travis build.

protomors:
Reason why build is broken now is because recently travis switched workers to Ubuntu Xenial by default. But Trusty is still available so for now I simply specified it in the config file.
Ultimately it will be better to adapt to recent distribution because support period of Trusty is ended. I will try it later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os: linux
+dist: trusty
+
 language: android
 
 jdk:


protomors:
It still fails. But looks like compilation errors were introduced in the latest commit to master https://github.com/Freeyourgadget/Gadgetbridge/commit/8449e55f9e9c174a27f67a52cf0cbfce7ab8e765
Before it worked fine https://travis-ci.org/protomors/Gadgetbridge/builds/563902442
Is this simply work in progress, or it compiles fine locally?

ashimokawa:
@protomors 
Yes I screwed up, thanks ;)
trusty to the rescue :+1:

PR #1572, opened 2019-07-26
imported R class

dakhnod:
without the import i am getting the Error that the class is unknown


ashimokawa:
@dakhnod 
Oh sorry, I already fixed that without seeing your PR

PR #1578, opened 2019-07-31
Amazfit Bip: Add emoji support when using custom font firmware

Nephiel:
This enables Gadgetbridge to display emoji on the Amazfit Bip, like other apps such as **Alert Bridge** or **Amazfit Tools**, if the user has flashed a custom font firmware with emoji support on the Bip.

A preference checkbox under `Settings` -> `Language and region settings` -> `Use custom font` toggles the feature.

Needs work, I wouldn't merge it as is:
* This preference needs to be settable per-device. As is, it is a global pref, and may cause other devices (i.e. Amazfit Bips without a custom font, or any other device) to display garbage text as long as this preference is enabled.
* Translations.


Nephiel:
* The preference to use a custom font is now per-device.
* AFAIK only translations are missing.

spiff42:
@Nephiel , great work! I would really like to try it out. I already have an emoji-font on my Amazfit bip, and I am using only gadgetbridge (installed via F-droid). Is there a way to easily try your version? An APK perhaps?

Nephiel:
> @Nephiel , great work! I would really like to try it out. I already have an emoji-font on my Amazfit bip, and I am using only gadgetbridge (installed via F-droid). Is there a way to easily try your version? An APK perhaps?

Sure, [here](https://storage.googleapis.com/gcs.nephiel.net/android/packages/GadgetBridge-0.35.2-usecustomfont.apk)'s my build of 35.2 if anyone wants to try it (at their own peril, of course, signatures won't match upstream releases so you'll have to reinstall to try it).

spiff42:
Happy to report that it works perfectly with Danish special characters as well as emoticons:
DSC_0553
I had to install [this font](https://forum.xda-developers.com/smartwatch/amazfit/amazfit-bip-cor-eu-support-font-pack-t3872210) for both to work.
Hope this gets merged soon!

ashimokawa:
@Nephiel 
Thanks for your contribution?
Would you mind if I merge this and then move all the code around? :smirk: 

Things I would change (although not yet clear how):

- I do not like the switch and if and cast in DeviceCommunicationService, that should not know anything about devices. There should be an abstract method somewhere else which does device specific text filtering and transliteration (here for bip and cor)
- The method that patches characters in a bip/cor specific way to custom locations in the bip font should not be in StringUtils as it is very specific and not generic.
- I would remove supportsCustomFont(), since all device that have updatable fonts "support" custom fonts, I mean patched fonts with extra characters etc. Instead the method mentioned above which does " device specific text filtering" should just do nothing or device to patch emoji if the config option is set
- The setting should not be added to the bip xml but should be a new one and added to the list of supported settings in Cor and Bip coordinators.

Nephiel:
Sure! Feel free to merge and move things as you see fit.

ashimokawa:
Anyone knows if there is an emoji font without japanese stripped out?

ashimokawa:
@Nephiel 
I have rebased and merged that, thanks!
(Will move the code around before the next release ;))

PR #1587, opened 2019-08-04
Use Rebble Store URI for Pebble watch app lookup

KonradIT:
Uses the rebble store as lookup for apps and watchfaces. Android-store-app will also intercept this link.


ashimokawa:
Thanks, merged by rebase

PR #1596, opened 2019-08-11
Add dedicated charts section into Settings

vanous:
This is a small PR which adds "Charts Settings" preference category into the menu. The "Enable left/right swipe..." goes there. In the future, i would like to add "Show weekly average" into to this settings - preference category.

There is an existing "Chart settings", which contains Min/Max Heart Rate but i am not sure if they do belong to the "Charts Settings". I did include it, but please let me know if i should remove it.


ashimokawa:
Thanks, merged by rebase.

PR #1597, opened 2019-08-11
Add Averages to charts.

vanous:
This adds Averages to Sleep/Steps, can be disabled in menu.

I am new to this. Please review and let me know how to improve this PR. I am sure something will be _fundamentally_ wrong from some point of view :)

thank you very much


ashimokawa:
Does not look fundamentally wrong ;)
I am not the author of the initial chart code, but I think what you did is in line with what was already there. Thanks!

I merged this by rebasing, so closing this manually

Issue #1598, opened 2019-08-11
MIME missing when sharing single GPX file from Activities screen


vanous:
#### Before reporting a bug, please confirm the following:
- [X] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [X] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [X] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:

When sharing single file from Activities, MIME is not specified, therefore some possible application/recipients are not listed. Not sure if MIME is 100% required or not - i understand that it must be there as per: https://developer.android.com/reference/android/content/Intent.html#ACTION_CREATE_DOCUMENT but my understanding might be lacking some details.

Captured intent:

```
intent://nodomain.freeyourgadget.gadgetbridge.screenshot_provider/external_files/Android/data/nodomain.freeyourgadget.gadgetbridge/files/gadgetbridge-walking-2019-07-28T19_25_05%2B02_00.gpx#Intent;scheme=content;launchFlags=0x3000000;end 
------------ 
ACTION: android.intent.action.VIEW 
DATA: content://nodomain.freeyourgadget.gadgetbridge.screenshot_provider/external_files/Android/data/nodomain.freeyourgadget.gadgetbridge/files/gadgetbridge-walking-2019-07-28T19_25_05%2B02_00.gpx 
URI: intent://nodomain.freeyourgadget.gadgetbridge.screenshot_provider/external_files/Android/data/nodomain.freeyourgadget.gadgetbridge/files/gadgetbridge-walking-2019-07-28T19_25_05%2B02_00.gpx#Intent;scheme=content;launchFlags=0x3000000;end 
FLAGS: 
FLAG_ACTIVITY_FORWARD_RESULT 
FLAG_ACTIVITY_PREVIOUS_IS_TOP 

------------ 
MATCHING ACTIVITIES: 
c:geo (cgeo.geocaching - cgeo.geocaching.CacheListActivity) 
Nova Launcher (com.teslacoilsw.launcher - com.teslacoilsw.launcher.RestoreBackupFileHandler) 
[...]

PR #1599, opened 2019-08-11
#1598 Added type to intent when sharing

vanous:
This is addressing #1598 . I have also tried to fix the `ActivitySummariesActivity.shareMultiple` but haven't found a good combination yet. What was interesting is that when in `AndroidUtils.viewFile` i used 
```
intent.setData(...
intent.setType(...
```
it didn't work, but setting both at the same time:
```
intent.setDataAndType(contentUri,"application/gpx+xml");
```
worked. Perhaps `ActivitySummariesActivity.shareMultiple` has similar issue, not sure.

Anyways, this one is solved. Please review and merge. I tested it successfully here. Thank you


ashimokawa:
Thanks, merged by rebase.

PR #1600, opened 2019-08-12
Add optional 30days charts view

vanous:
Actually a simple modification, which adds an options into Settings to change Charts **Steps per Week** and **Sleep per Week** into **Steps per Month** and **Sleep per Month**. 

Default is false, to keep traditional view.

When in Month view, Charts label, instead of short day name, uses day_of_month value, for better visibility.

Monthly charts seem quite dense at first due to axis values, but as they can be zoomed in, these values are actually very useful. This could perhaps be improved in the future by showing less density when zoomed in (like in the labels).


vanous:
This is ready, if you like it.

ashimokawa:
Thanks ;) - rebased and merged

PR #1604, opened 2019-08-16
Add Status and Alarms widget

vanous:
This is partially based on the Sleep Widget and on the way the charts retrieve and analyze activities data.

- provides a widget, which upon clicking gets the samples from the database (it goes through all devices in GB and fetches current day activity samples)
- widget auto-refreshes every six hours (21600000ms)
- the activity analyses is copy/paste from charts package, but some methods are marked private and protected and i didn't want to break anything there, so i duplicated the class completely into the model package.

I have done as much due diligence as possible. What i am not sure about is me passing the context, perhaps that is not necessary, please let me know.


vanous:
I am still working on this. Please feel free to comment at any time.

vanous:
Hi @ashimokawa 

I am not planning to chase up the upcoming weekend release, especially since you might have some code comments (please do if needed), but accidentally i have finished today.

This provides new widget with status and alarms. Features:

- steps and sleep for the current day (DailyTotals)
- state of the device (connecting/disconnected) or battery % if connected
- quick access to GB main screen (tap GB icon)
- quick access to Charts activity (tap steps/sleep)
- quick data refresh (header/tap device name)
- popup menu to set an alarm with multiple (fixed) presets available + daily sleep goal as one of the options. This is based on the existing widget - overwriting the first available alarm slot of the device.


Comments:
- i had to add one broadcast intent call to AbstractFetchOperation/handleActivityFetchFinish to receive info about operationFinished after data fetch, as the widget only gets data based on intent calls, rather then time based polling.
- i also listen to nodomain.freeyourgadget.gadgetbridge.gbdevice.action.device_changed to get the device state
- AppWidgetProvider doesn't have appcompat so png icons are required, rather then vectors. I only used existing icons, but added preview png image for the widget selection screen.
- tested with Bip, on Android O (8.0)
- possible future improvements: add "remove alarm item to alarms popup" by adding new/modifying createSingleShot method in AlarmUtils with the possibility to set enable true/false.
- it took me several nights to discover and plow through all i needed, so the log is quite extensive and might need to be squashed

As before - i might have done some pretty crazy mistake somewhere. I hope i haven't, but kindly please check.

ashimokawa:
@vanous 
Thank you for your contribution, it really much appreciated!
Unfortunately I did not have the time for a review and I will tag 0.36.0 today and then review and probably land this into 0.36.1 together with #1578 

There are some important fixes which have accumulated and I do not want to push the release any further.

vanous:
Thank you too :)

ashimokawa:
@vanous
Is there a reason why you did not extend AbstractGBActivity (which takes care of theme and language changes) and used "Activity" instead?

And is the only reason you copy/pasted the analysis class the method visibility? (if so you can change that)

vanous:
OK, will look at it and change as needed.

vanous:
Alright, so i have:

- changed the ActivityAnalysis and now use the .charts version

- have deeply looked into the WidgetAlarmsActivity and tried to make it extend AbstractGBActivity. My issue there is, that somehow the themes collide and manifest based @style/Theme.AppCompat.Light.Dialog for the popup is not applied or is overextended with the non-dialog theme (i tried to add work this out in the AbstractGBActivity.init etc..., to set custom this dialog based theme during init, but no luck at this point, it always opens full screen rather then popup. 
I went back and extend the popup on Activity but have checked that language is switching correctly and i have also based all text coloring in the widget to already defined colors (like @color/primary_dark...) to make the look consistent and globally adjustable.

Lastly i struggled with rebasing :/ but i should be back not conflicting so i hope this doesn't cause even more troubles.

ashimokawa:
@vanous 
Thanks! I hope I can merge this tonight!

ashimokawa:
Squashed and rebased while preserving contribution ownership of @vanous 

Thanks! :)

PR #1610, opened 2019-08-24
adds custom renderer for better view of many columns

vanous:
When 30days charts view is used, angle the barchart labels by -90°.


ashimokawa:
thanks :)

PR #1626, opened 2019-09-04
remove preference keys when deleting device

vanous:
I cannot really test this well without another device, will probably test in next few days.


vanous:
tested, works.

ashimokawa:
thanks, rebased

PR #1638, opened 2019-09-13
Add further ZeTime support

Lightwars:
- I just added a method, which sets the language accordingly to the phones language preferences if supported otherwise english is used on the watch.

- Enable support for unicode emojis: the watch supports a limited selection of emojis, but we want to use them.


ashimokawa:
Thanks, merged by rebase rebased

PR #1641, opened 2019-09-15
Adds plurals to a string of the Widget.

### LINKED ISSUE #1637: Plural forms
In Hebrew there is a dual form for hours, so we have 1, 2, >2, which are a singular form, a dual form and a plural form.

%d hours is wrong in our case (and also in several other languages such as Arabic and Russian) because we need a different case according to the amount of hours.

https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/res/values/strings.xml#L741

https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/res/values/strings.xml#L743

vanous:
This fixes #1637.


ashimokawa:
Thanks, rebased and merged

PR #1643, opened 2019-09-16
Refresh control center on menu change

vanous:
Sorry for these small changes, i will try to get better in the future.


ashimokawa:
@vanous 
What does this fix?
Is it necessary to recreate the whole Activity?

vanous:
The fob doesn't get drawn, if it was enabled after being previously disabled. Can I redraw/recreate just the fob?

ashimokawa:
@vanous 
Yes, move the fab local variable to a private class member variable then just call fab.show()/fab.hide()

vanous:
Thank you very much, i will look into it.

ashimokawa:
rebased, squashed and merged - will do some further changes

Issue #1621, opened 2019-09-01
skip small sleep times


FransM:
Today I noticed my Mi Band 4 recorded a short sleep while I was not sleeping but actually I was just reading the paper at the kitchen table.  I'm not sure if this was related to the way I wore the band, but this shows up as the new sleep end time in "your sleep".
It might be better to filter out sort sleeps (as false reading).
And, if there are more sleep intervals (e.g. an afternoon nap", it may be better to say:
"You slept from xx:xx to yy:yy and from aa:aa to bb:bb"

Your wearable device is:

Mi Band 4 FW 1.0.5.66
Your android version is:

7.1.1
Your Gadgetbridge version is:

0.36

tomcat-xx:
Did you activate "Use heart rate sensor to improve sleep detection"? I have, but cannot say if it helps. But maybe different for different users.

As I wrote in another issue I had my first deep sleep phase in office. Until now this did not happen again with above option activated. And luckily I'm using GB and not Mi Fit with direct connection to my boss through the cloud ;-)

maurosandes:
I have the same problem Mi band 4 FW 1.0.6.16

FransM:
Fix for issue #1646 ignores sleeps < 5 minutes. Seems a fair solution. 
Leaving this open for now as I would like to get feedback if the ignore time should be a configurable option

vanous:
I would close thi,  in my opinion. If the need comes up, open a new issue.

IzzySoft:
I agree with @vanous here, with a slight deviation: if the need arises, we can also reopen this one. Thanks everybody!

Issue #1620, opened 2019-09-01
odd sleep data report


FransM:
### I got Gadgetbridge from:
* [x ] F-Droid
* [ ] I built it myself from source code (specify tag / commit)

#### Your issue is:
I noticed something odd today.
Yesterday I woke up at 8.40.
This morning I woke up at 8.16 and checked the "my sleep" tab at 8:25
It told me I slept from 8:25 until 8.16 This looks very odd.
See attached screenshot.

![Screenshot_20190901-082535](https://user-images.githubusercontent.com/1893299/64081581-7ac77a00-cd03-11e9-9a79-bb1b515a1df3.png)

Apparently what happened is that the last part of my sleop of the day before was taken as the start of the sleep.

Maybe we should say that the start of sleep is always after 12.00 mid day (or make this a config item)

#### Your wearable device is:

Mi Band 4 FW 1.0.5.66

#### Your android version is:
7.1.1

#### Your Gadgetbridge version is:
0.36

tomcat-xx:
Same for me. But when I remember right, later same day it was displayed correctly.

But anyway sleep recognition with bands is not really good. For me I think there should be deep sleep phases ... in the night ... as my first recorded one was in the office ;-)

FransM:
I feel issue #1646 fixes the issue as the report does not cause the confusion reported in this ticket. As submitter of the ticket I am closing it.

vanous:
@FransM also, the https://github.com/Freeyourgadget/Gadgetbridge/pull/1677 helps to limit the sleep time a bit. Can you please close this now, i think you forgot about it... :)
cheers

FransM:
Will look at it tomorrow

Groetjes, Frans 
 
  Op za, okt. 12, 2019 om 18:16 schreef vanous<notifications@github.com>:   
@FransM also, the #1677 helps to limit the sleep time a bit. Can you please close this now, i think you forgot about it... :)
cheers

FransM:
Thought I already closed this, not sure why that did not come through. Closing again

Issue #1277, opened 2018-09-21
New sleep start/end times feature is overeager


LuccoJ:
![screenshot_gadgetbridge_20180921-162959](https://user-images.githubusercontent.com/1684791/45888532-95d4fe80-bdbe-11e8-88c4-da12f70c7ac1.png)

In short since I'm typing and walking, I think this feature should not count several hours of awake time after a short nap (or in this case, a misdetectrd nap) as just an awakening during sleep, resulting in the idea that I sleep from 17:00.

I think this should be solved by considering only awakenings briefer than about an hour, or half the time of the preceding sleep (whichever is shorter) as being "part of" the sleep session.

cpfeiffer:
@elagin Do you have an idea how to fix this? In theory one can have many separate sleep periods during a day. One could take the longest period. A sleep period could still be interrupted by wakeups (short enough to be tolerable). Maybe we need some more structure ini the ActivityAnalysis classes to express this.

LuccoJ:
By the way, I'm not sure if it's worth opening a separate issue, but I've thought for a while that the way total slept time is computed should probably be rethought: typically, I look at my asleep time after waking up, but if I've waken up earlier than I did the previous day, it will consider part of the previous day's sleep in the calculation, resulting in an overestimate of the night's slept time. This estimate will then go down later during the day, which is a bit counterintuitive.

So perhaps, aside from the sleep start/end time, the total slept time should only consider the past 12 hours, or if that's not appropriate because it'd miss naps, at least the past 20 hours? Something that makes reasonably sure the previous day's sleep isn't tacked in, at any rate.

PR #1646, opened 2019-09-18
Multiple sleep sessions per day

Q-er:
Added a feature to detect seperate sessions of sleep in one day, e.g. a night time sleep and a midday nap.

To seperate sessions, you need to be awake for more than 2 hours.
So if you wake up only for a few minutes, it still counts as 1 session.
Awake time during a session does not increase total sleep time. (like before)

Not a fix, but related to #1620 as it would show 2 sleep sessions: 8:25 to 8:50 AND 0:30 to 8:16. (also #1277)
Related to #1621 as it ignores "sessions" shorter than 5 mins.

**Screenshots**

Before

After


ashimokawa:
Thanks, rebased and merged

Issue #1636, opened 2019-09-10
gadgetbridge crashes when attempting to open amazfit bip activities 


camelCaseNick:
### I got Gadgetbridge from:
* [x] F-Droid
* [ ] I built it myself from source code (specify tag / commit)

#### Your issue is:
When trying to open an activity from the activities list for gadgetbridge crashes.

See: https://imgur.com/a/KSB90TI

#### Your wearable device is:
amazfit bip v1.1.5.56

#### Your android version is:
Android 9

#### Your Gadgetbridge version is:
0.36.0

petrvanekrobe:
Most probably I cannot help you very much with this, but perhaps try to provide some evidence/log files, please make sure to read this first: https://github.com/Freeyourgadget/Gadgetbridge/wiki/Log-Files

vanous:
Most probably I cannot help you very much with this, but perhaps try to provide some evidence/log files, please make sure to read this first: https://github.com/Freeyourgadget/Gadgetbridge/wiki/Log-Files

camelCaseNick:
@vanous , sorry, I did not allow OsmAnd to be opend from a different app (only main intent from home screen). Sorry, for wasting your time with my configuration problems.

PR #1651, opened 2019-09-24
Show toast in case no app is installed which can handle GPX files

Low012:
As mentioned in https://github.com/Freeyourgadget/Gadgetbridge/issues/1636, the Gadgetbridge app crashes if no other app is installed which can handle GPX files and the user tries to open a GPX file anyway. The issue was closed, but in my opinion the situation should be handled more gracefully than letting the app crash without telling the user the reason for this behaviour.

With the changes in this PR the user will see a recommendation to install an app which can handle GPX files and the Gadgetbridge app will not crash.


ashimokawa:
@Low012 
Thanks, looks good to me. Will test and merge before the next release (0.36.2 is already done and we are waiting for F-Droid)

Also that you for being almost the only person who can spell Gadgetbridge correctly without camel case. :+1:

ashimokawa:
rebased and merged, thanks ;)

PR #1658, opened 2019-09-28
Basic Makibes HR3 support

Cre3per:
Supports reboot, factory reset, set date/time, alarms, notifications, find device.
Doesn't support any operations that require reading from the watch.
There are a couple more commands in MakibesHR3Constants.java which I haven't implemented.
My watch died, I won't be able to continue working on this unless I manage to fix it.


ashimokawa:
@Cre3per 
Thanks I have rebased and mergend this.
About the preferences, they should be done per-device now and not clutter the settings menu (there will be a settings icon on the device card if the device support per-device settings).

I will move the settings.

PR #1661, opened 2019-10-02
Amazfit bip: Support flashing latest latin font

MyTimeKill:


ashimokawa:
Did you try? ;)

MyTimeKill:
Yes, it works

PR #1670, opened 2019-10-09
Implemented Makibes HR3 Features

Cre3per:
I fixed my watch and added some more features, see changelog.
Thanks to @ashimokawa for improving my code last time, I'm not a Java dev.


ashimokawa:
@Cre3per 
Cool, thanks.

Just one note: There is support for what you call "reverse find" in Gadgetbridge already, even an extra activity which makes the phone ring loud. You could add your vibration code there if you like but it would be cool if your code just does it the Gadgetbridge way instead of reinventing the wheel ;)
Look at HuamiSupport.java and search for GBDeviceEventFindPhone and references

Cre3per:
@ashimokawa Thanks, I didn't see that. I added vibration to `FindPhoneActivity` and now use it for the Makibes HR3. I also added `devicesettings_find_phone.xml` to enable/disable/timeout the activity so that devices which don't pair can disable it to prevent malicious devices from starting the activity.

ashimokawa:
Would you mind if I squash all changes into one commit (your authorship will be preserved) ? Or would you like to have all commits preserved, most have no real description so....

Cre3per:
Do as you please, I don't mind

ashimokawa:
rebased and merged, did not squash, saw you had real changelogs ;)

Issue #1668, opened 2019-10-07
"Your sleep" charts time range


vanous:
#### Before requesting a new feature, please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

Right now, the "Your sleep" charts time span is "now - 24h". This creates several issues, for example the new addition for multiple sleep sessions (thanks to @Q-er for that!) shows me that if i look at my last night sleep in the morning, i will see some sleep sessions from last night + some hours from the night before. Secondly, the "Your sleep" chart is basically just a copy of the "Activity and sleep" charts, meaning - it is redundant and taking screen space. 

My suggestion is to change the time range for the "Your sleep" tab, where we would basically do a time range:

from: first detected sleep in "midnight -12h" (or "midnight - 8h" to eliminate afternoon nap...)
to: last sleep session's end. 

So the "Your sleep" chart would actually only show sleep, rather then "activity + sleep".

What do you think?

vanous:
Nobody opposed this so i went ahead as per the PR to make it look as i described:

<img src="https://user-images.githubusercontent.com/3680926/66700725-34970a80-ecf4-11e9-9954-3c58fe7092e4.jpg" width="400">

vanous:
Fixed by  #1677

PR #1677, opened 2019-10-12
Limit sleep charts range

vanous:
Addressing https://github.com/Freeyourgadget/Gadgetbridge/issues/1668 . 
This limits the "Your sleep" charts range to start with first detected sleep start and end with last detected sleep end. Possible improvement: Scale up, but need more samples due to possible need for HR measurements to be drawn correctly.


vanous:
@ashimokawa this is a pretty small change to the charts. It limits the data for sleep charts from 13:00 to 13:00, rather then using "now" - 24h as the range, which ended up selecting last night sleep. But the results seems very pleasing to me.

ashimokawa:
@vanous
Ok sorry I just tagged. :/

ashimokawa:
Thanks, rebased and merged

PR #1679, opened 2019-10-13
Allow apps to update existing notifications

M66B:
An app like [FairEmail](https://github.com/M66B/FairEmail/) first fetches the header of a message, displays a new message notification and later fetches the message text and updates the existing new message notification with the fetched message text. The latter might not happen if the message is large and the connection is metered (mobile or paid WiFi).

The minor change in this pull request makes updating the notification possible.


M66B:
This is not good enough because the existing notification won't be removed.

PR #1688, opened 2019-10-16
Makibes HR3 fixes

Cre3per:
The Makibes HR3 coordinator matches the band by name. The english version of the band has a different name than the chinese version. There's another version of the band that doesn't show its name in the menu, so I can't add it without having one at hand, that one might not work.

The connection used to time out after bluetooth was disabled on both devices. Pairing/Bonding fixed it.


ashimokawa:
Cool. Will merge tonight.

ashimokawa:
Thanks, rebased and merged

PR #1693, opened 2019-10-19
Fixed notification remove callback by looking up the original id

dakhnod:


ashimokawa:
@dakhnod 
Interesting. Thanks.
I have to look up the original reason why I did that - it was something really bad.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/externalevents/NotificationListener.java
```diff
@@ -638,17 +641,14 @@ public void onNotificationRemoved(StatusBarNotification sbn) {
...
             }
         }
-        // FIXME: DISABLED for now
-        /*
         if (shouldIgnore(sbn))
             return;
 
         Prefs prefs = GBApplication.getPrefs();
-        if (prefs.getBoolean("autoremove_notifications", false)) {
+        if (prefs.getBoolean("autoremove_notifications", true)) {
```
ashimokawa: Why change the default?

dakhnod:
The first time I tried to uncomment the code because of a long to int conversion of getPostTime().

I changed the default because AFAIK the option from the settings is also disabled...

PR #1700, opened 2019-10-23
Mi Band 4: Enable Emoji Font setting

HardLight:
There is custom emoji font for Mi Band 4 at myamazfit.ru. We should enable custom font setting.


ashimokawa:
Thanks, was unaware of the font

PR #1703, opened 2019-10-24
Add alternative color to heartrate in charts

vanous:
Make HR drawing better for accessibility.


ashimokawa:
Thanks, rebased and merged

Issue #1705, opened 2019-10-25
improve sleep time reporting


FransM:
#### Before requesting a new feature, please confirm the following:
- [x] I have read the [wiki](https://github.com/Freeyourgadget/Gadgetbridge/wiki), and I didn't find a solution to my problem / an answer to my question.
- [x] I have searched the [issues](https://github.com/Freeyourgadget/Gadgetbridge/issues), and I didn't find a solution to my problem / an answer to my question.
- [x] If you upload an image or other content, please make sure you have read and understood the [github policies and terms of services](https://help.github.com/articles/github-terms-of-service/#1-responsibility-for-user-generated-content)

#### Your issue is:
This is on the My Sleep page and the 24hr window.
If I slept yesterday from say 11 pm to 7 am, and last night from 11 pm to 6 am and check my sleep at 6.15 am it says I slept from 
06:15 to 07:00
23.00 to 06.00
5 minutes later it will report
6.20 to 7:00
23.00 to 6:00

As the start date of the first sleep period is on the day before I suggest to report this as:
.... to 7:00
23:00 to 06:00
So a start time before the 24 hr window is shown as ellipsis

#### Your wearable device is:
mi band 4
*Please specify model and firmware version if possible*

#### Your android version is:
71

#### Your Gadgetbridge version is:

latest from fdroid

*New requests about already solved/documented topics could be closed without further comments. Same for too generic or incomplete reports.*

IzzySoft:
That "works as designed": this tab shows a "moving window" of the last 24 hours. So 24h before 6:15 it was… well, 6:15, right? So that's the earliest reference point. 5min later it's 6:20, so that becomes the earliest reference point.

vanous:
This is actually fixed in an upcoming release. I added a split  at 13:00 hours, so both the charts and sleep sessions only show last night (plus maybe siesta) sleep time. We can close this. 
:blush:

IzzySoft:
Uh? Not arguing but wondering: will night-shift workers now complain, because they start their sleep sessions around 10am? A rolling 24h window is easy to grasp. A break at a "random point" is not. Why 1pm – and not e.g. 12am or 2pm? Or configurable? Or keep the "rolling window"? Or break at midnight to fix it to the date? :confused:

vanous:
> Uh? Not arguing but wondering: will night-shift workers now complain, because they start their sleep sessions around 10am? 

[...]

PR #1709, opened 2019-10-26
Add switch to set old/new style of sleep detection range

vanous:
Addressing concern for night workers and those who prefer rolling 24h window. Also, changed split time to 12, as it is easier to explain and describe (Noon time).


vanous:
https://github.com/Freeyourgadget/Gadgetbridge/issues/1705

PR #1713, opened 2019-10-30
Add Set Activity Fetch Time to debug to set lastSyncTimeMillis

vanous:
This allows to go back in time to re-fetch data if needed.
I auto re-formatted the DebugActivity file. Generally, i try not to do this but i left it this time, please tell me if this is not good.


ashimokawa:
@vanous 
Will test, I had problems with the layout on small devices. So I am cautious.

vanous:
I see. What is the best way for me, setup emulator with smaller resolution /screen? Do you remember the size?

ashimokawa:
You dont need the emulator, you can use the preview in android studio and change the emulated

vanous:
OK. Seems OK here, tried bunch (probably pretty much all sub 5") of devices from Generic Phones.

ashimokawa:
rebased and merged, thanks!

PR #1716, opened 2019-11-01
Add forward/backward buttons to charts

vanous:
Allows to move faster in larger data sets.


ashimokawa:
Thanks! Rebased and merged!

PR #1743, opened 2019-12-09
Support for Bangle.js (and other Nordic UART devices)

gfwilliams:
This PR adds support for Espruino [Bangle.js](https://www.espruino.com/Bangle.js). It connects via BLE using the Nordic UART service and sends/receives JSON for common notifications/etc.

Currently handles notifications,calls,music - but doesn't receive/store health data yet.

The Bangle.js half of the implementation is at https://github.com/espruino/BangleApps/blob/master/apps/gbridge/widget.js

I'll get some proper docs sorted for this later on as well.


app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/banglejs/BangleJSCoordinator.java
```diff
@@ -0,0 +1,150 @@
...
+              name.startsWith("MDBT42Q") ||
+              name.startsWith("Espruino"))) 
+            return DeviceType.BANGLEJS;
+
+        return DeviceType.UNKNOWN;
+    }
+
+    @Override
+    public int getBondingStyle(){
+        return BONDING_STYLE_NONE;
```
ashimokawa: BONDING_STYLE_NONE means that the device should not be paired the bluetooth level for BLE devices chose to do some custom key exchange on a higher layer. Did you have that in mind?

ashimokawa:
@gfwilliams 
Thanks Gordon!
I will merge this this evening.

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/banglejs/BangleJSCoordinator.java
gfwilliams: I assumed that it was such that if Bangle.js specifically requested bonding you'd have it, but if not then you wouldn't? As Bangle.js does have the ability to do bonding I think you're right, BONDING_STYLE_ASK would probably be preferable.

ashimokawa: BONDING_STYLE_ASK is currently not used by any device AFAIK, hope it works ;)

So bangle.js has zero security and anyone could connect?

ashimokawa:
Merged! Thanks!

app/src/main/java/nodomain/freeyourgadget/gadgetbridge/devices/banglejs/BangleJSCoordinator.java
gfwilliams: Ahh, sorry. Seems to work though. I saw the comments and they said it was preferable to use BONDING_STYLE_ASK :)

By default there's zero security, yes. There's a bunch of stuff it can do (including whitelisting) but it's not enabled by default now. At the start it used to be, but I was getting so many support requests like "I paired this with A and now it's broken" that I had to disable it :)

